PR #75, opened 2026-04-10
fix(web): handle Tooltip formatter value to fix TypeScript build error

### LINKED ISSUE #74: Docker build fails due to TypeScript error in dimension-radar.tsx
when following the official setup (docker compose up --build), the frontend build fails during next build due to a TypeScript error in web/src/components/dashboard/dimension-radar.tsx (around line 40)

the issue is caused by the Tooltip formatter being typed as (value: number) while the underlying Recharts type allows value to be number | undefined (and potentially other types). This causes a strict type check failure during the production build:

```
Type error: Type '(value: number) => ...' is not assignable...
Type 'undefined' is not assignable to type 'number'
```
As a result the Docker setup cannot complete and the project fails to start

Note: The app may still run in development mode (pnpm dev) since type errors are not enforced there, but the production build (used in Docker) is currently broken

A simple fix is to handle non-number values safely in the formatter (e.g. checking typeof value === "number" before using it)

Kaushik-Kumar-CEG:
this fixes #74 

Resolves a TypeScript error in the Tooltip formatter in `dimension-radar.tsx` that caused `next build` (and Docker setup) to fail
The formatter previously assumed the value was always a number. This adds a simple type check to handle it safely

This allows `docker compose up --build` to run without errors. Small change, but it ensures the project builds correctly from a clean setup

### DIFF
--- a/web/src/components/dashboard/dimension-radar.tsx
+++ b/web/src/components/dashboard/dimension-radar.tsx
@@ -37,7 +37,7 @@ export function DimensionRadar({ dimensionScores }: DimensionRadarProps) {
         <PolarGrid />
         <PolarAngleAxis dataKey="dimension" tick={{ fontSize: 12 }} />
         <PolarRadiusAxis angle={90} domain={[0, 100]} tick={{ fontSize: 10 }} />
-        <Tooltip formatter={(value: number) => [`${value.toFixed(0)}/100`, "Score"]} />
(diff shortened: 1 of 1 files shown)


Kaushik-Kumar-CEG:
before fix:

image


after fix:

image

Haz3-jolt (approved):
Thanks

Issue #167, opened 2026-04-12
Support short/prefix ID matching for review and eval commands (like git short SHA)


dhanpraja231:
## Problem

Currently, commands like `observal admin review show <id>`, `approve`, `reject`, and eval commands require the full UUID to be provided. This is tedious, especially since `review list` truncates IDs anyway (see #166).

## Proposal

Allow prefix matching on IDs, similar to how `git checkout` accepts a short SHA prefix. For example:

```bash
# Instead of the full UUID:
observal admin review show 8a3f29b1-4c7e-4d2a-9f1b-abc123def456

# Accept a unique prefix:
observal admin review show 8a3f
```

This should apply to all commands that accept an ID:

- `observal admin review show <id>`
- `observal admin review approve <id>`
- `observal admin review reject <id> --reason "..."`
- `observal admin eval run <agent-id>`
- `observal admin eval scorecards <agent-id>`
- `observal admin eval show <scorecard-id>`
- `observal admin eval compare <agent-id>`
- `observal admin eval aggregate <agent-id>`

## Behavior

- If the prefix matches exactly one entry, use it.
- If the prefix matches multiple entries, return an error listing the ambiguous matches.
- Full UUIDs should continue to work as-is.

Kaushik-Kumar-CEG:
I'd like to take on this issue.I'll build a serverside prefix resolver and update the API/CLI to accept both short IDs and full UUIDs. Could you assign it to me?

Haz3-jolt:
Already implemented. config.resolve_alias() supports prefix ID matching, row number references, and @alias lookups across all CLI commands.

PR #177, opened 2026-04-12
feat: support short prefix ID matching for review and eval commands (#167)

Kaushik-Kumar-CEG:
## Summary
Allows users to use short ID prefixes instead of full UUIDs, similar to `git` short SHAs.

```bash
# Before
observal admin review show 8a3f29b1-4c7e-4d2a-9f1b-abc123def456

# After
observal admin review show 8a3f
```

## Changes

**`api/deps.py`** — Added `resolve_prefix_id` helper:
- Normalizes input, tries exact UUID first, then prefix match via `cast(model.id, String).like(...)`
- Rejects prefixes shorter than 4 chars
- Returns 400 with match list if prefix is ambiguous
- Supports `extra_conditions` and `load_options` for route-specific needs

**`api/routes/review.py`** — Updated `_find_listing` to use the helper across all 5 component models with cross-model ambiguity detection.

**`api/routes/eval.py`** — Changed path params from `uuid.UUID` to `str`, all lookups use the helper.

**`api/routes/agent.py`** — Replaced `_agent_id_clause` with `_load_agent` which resolves by UUID, prefix, or name with eager loading.

**CLI** — No changes required. All ID arguments were already typed as `str` and pass through correctly.

## Behavior
- Unique prefix → resolves to matching record
- Ambiguous prefix → 400 error listing up to 5 matches
- Prefix < 4 chars → 400 error
- Full UUIDs → continue to work unchanged
- Name-based lookups → continue to work unchanged

### DIFF
--- a/observal-server/api/deps.py
+++ b/observal-server/api/deps.py
@@ -3,7 +3,7 @@ from collections.abc import AsyncGenerator
 from functools import wraps
 
 from fastapi import Depends, Header, HTTPException
-from sqlalchemy import select
+from sqlalchemy import select, cast, String
 from sqlalchemy.ext.asyncio import AsyncSession
 
 from database import async_session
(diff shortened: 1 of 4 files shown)


Haz3-jolt:
Can you sign off your commits. It's a -s flag.

PR #368, opened 2026-04-18
Kiro agent pull for Windows/Powershell

naraen-ram:
**Problem**
`observal pull --ide kiro` generates agent configs with Unix-only shell commands and a hardcoded Linux DB path. None of this works on Windows/PowerShell, so the entire Kiro hook telemetry pipeline was broken for Windows users. On top of that, the `observal-shim` couldn't spawn MCP servers on Windows because `asyncio.create_subprocess_exec` can't find `.cmd` executables and `connect_read_pipe/connect_write_pipe` don't work on the Windows Proactor event loop.

**Fix**
The CLI now sends `sys.platform` to the server via a new optional platform field on the install request. When platform == "win32", the server generates PowerShell-compatible hook commands using `python -m observal_cli.hooks.kiro_hook --url ... --agent-name ...` instead of the Unix pipeline. The hook scripts now resolve the Kiro SQLite DB path at runtime (`%LOCALAPPDATA%` on Windows, `~/.local/share` on Unix). The shim uses `shutil.which()` to resolve `.cmd` executables and a thread-based stdin reader to replace the broken asyncio pipe handling on Windows. All changes are gated behind `sys.platform == "win32"` — macOS/Linux behavior is completely unchanged. The pull command should be used with `--dir ~` to write the agent file to the home directory where Kiro can find it.

**Extra fixes and additions**
`observal-shim` couldn't find `npx.cmd` on Windows — fixed with `shutil.which()`
`observal-shim` crashed on Windows due to asyncio pipe incompatibility — fixed with thread-based stdin relay and direct stdout writes
Cleaned up a duplicate `run_shim` function that was left behind during editing
Added 19 new tests covering Windows fix checking, Unix preservation, hook config generator, and platform-aware DB path resolution — all 126 tests pass

**Known issue** 
Traces don't show up in the Observal UI when using kiro-cli. After investigation, kiro-cli displays the hook commands in its UI panel but doesn't actually execute them when events fire. The hook commands are correct and work when run manually — kiro-cli just doesn't trigger them. Not exactly sure how to fix this...

### DIFF
--- a/magic-agent.json
+++ b/magic-agent.json
@@ -1,0 +1,28 @@
+{
+  "name": "magic-agent",
+  "version": "1.0.0",
+  "description": "Agent with 21st.dev Magic MCP for UI component generation",
+  "owner": "admin",
+  "prompt": "You are a UI development assistant. Use the magic MCP server to generate beautiful UI components.",
+  "model_name": "claude-sonnet-4",
+  "model_config_json": {},
+  "supported_ides": ["kiro"],
+  "mcp_server_ids": [],
+  "components": [],
+  "external_mcps": [
+    {
+      "name": "magic-mcp",
+      "command": "npx",
+      "args": ["-y", "@21st-dev/magic-mcp"]
+    }
+  ],
+  "goal_template": {
+    "description": "UI component generation",
+    "sections": [
+      {
+        "name": "Task",
+        "description": "What UI component to build"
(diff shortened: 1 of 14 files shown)


ShaanNarendran:
Merge conflicts, will have to rebase and resolve

PR #435, opened 2026-04-20
feat: add observal migrate telemetry commands for ClickHouse deep copy

naraen-ram:
Adds deep copy for the cloud migration: three new CLI commands that export, import, and validate ClickHouse telemetry data using Apache Parquet. Requires PostgreSQL shallow copy to be completed first. Also adds a sidecar manifest to shallow copy exports for it's coordination.

**Commands Added:**
1. `observal migrate export-telemetry` : Export all 6 ClickHouse telemetry tables to Parquet files
2. `observal migrate import-telemetry` : Import Parquet files into a target ClickHouse instance
3. `observal migrate validate-telemetry` : Verify checksums, compare row counts, and validate FK references against PostgreSQL

**Execution Example:**
```
uv tool install --editable . --force

# Phase 1 prerequisite (creates sidecar manifest for Phase 2)
observal migrate export --db-url "postgresql://..." --output test-export.tar.gz

# Export telemetry
observal migrate export-telemetry \
  --clickhouse-url "clickhouse://..." \
  --manifest test-export.manifest.json \
  --output-dir telemetry-export

# Validate checksums 
observal migrate validate-telemetry --input-dir telemetry-export

# FK validation against PostgreSQL 
observal migrate validate-telemetry --input-dir telemetry-export --target-db-url "postgresql://..."

# Row count comparison against ClickHouse 
observal migrate validate-telemetry --input-dir telemetry-export --clickhouse-url "clickhouse://..."

# Import into ClickHouse 
observal migrate import-telemetry --clickhouse-url "clickhouse://..." --input-dir telemetry-export

# Idempotent re-import 
observal migrate import-telemetry --clickhouse-url "clickhouse://..." --input-dir telemetry-export

# Post-import validation 
observal migrate validate-telemetry --input-dir telemetry-export --clickhouse-url "clickhouse://..."
```

### DIFF
--- a/observal_cli/cmd_migrate.py
+++ b/observal_cli/cmd_migrate.py
@@ -5,7 +5,9 @@ from __future__ import annotations
 import asyncio
 import hashlib
 import json
+import logging
 import os
+import re
 import shutil
 import tarfile
 import tempfile
@@ -24,6 +26,8 @@ from rich import print as rprint
 
 from observal_cli import client
 from observal_cli.render import spinner
+
+logging.getLogger("httpx").setLevel(logging.WARNING)
 
 # ── Constants ────────────────────────────────────────────
 
@@ -90,6 +94,27 @@ JSONB_COLUMNS: dict[str, list[str]] = {
     "exporter_configs": ["config"],
 }
 
+# ── Phase 2: ClickHouse telemetry constants ──────────────
+
+CLICKHOUSE_TABLES: list[dict[str, str | list[str]]] = [
+    {"name": "traces", "engine": "replacing", "time_col": "start_time", "fk_cols": ["agent_id", "mcp_id", "user_id"]},
+    {"name": "spans", "engine": "replacing", "time_col": "start_time", "fk_cols": ["agent_id", "mcp_id", "user_id"]},
+    {"name": "scores", "engine": "replacing", "time_col": "timestamp", "fk_cols": ["agent_id", "mcp_id", "user_id"]},
+    {"name": "audit_log", "engine": "mergetree", "time_col": "timestamp", "fk_cols": ["actor_id"]},
+    {"name": "mcp_tool_calls", "engine": "mergetree", "time_col": "timestamp", "fk_cols": ["mcp_server_id", "user_id"]},
+    {"name": "agent_interactions", "engine": "mergetree", "time_col": "timestamp", "fk_cols": ["agent_id", "user_id"]},
+]
+
+FK_PG_TABLE_MAP: dict[str, str] = {
+    "agent_id": "agents",
+    "mcp_id": "mcp_listings",
+    "mcp_server_id": "mcp_listings",
+    "user_id": "users",
+    "actor_id": "users",
+}
+
+EPOCH_SENTINELS: set[str | None] = {None, "", "1970-01-01 00:00:00.000", "1970-01-01 00:00:00"}
+
 
 # ── PGEncoder ────────────────────────────────────────────
 
@@ -145,6 +170,34 @@ class ValidationResult:
     cross_db_results: dict[str, tuple[int, int]] | None
 
 
+@dataclass
+class TelemetryExportResult:
+    output_dir: str
+    migration_id: str
+    table_results: dict[str, dict]
+    total_rows: int
+    total_size_bytes: int
+    duration_seconds: float
+
+
+@dataclass
+class TelemetryImportResult:
+    migration_id: str
+    tables_imported: int
+    tables_skipped: list[str]
+    rows_imported: dict[str, int]
+    duration_seconds: float
+    warnings: list[str]
+
+
+@dataclass
+class TelemetryValidationResult:
+    checksums_valid: bool
+    checksum_results: dict[str, bool]
+    fk_results: dict[str, list[str]] | None
+    row_count_results: dict[str, tuple[int, int]] | None
+
+
 # ── Helper functions ─────────────────────────────────────
 
 
@@ -184,6 +237,18 @@ def _sha256_file(path: Path) -> str:
         for chunk in iter(lambda: f.read(8192), b""):
             h.update(chunk)
     return h.hexdigest()
+
+
+def _parse_clickhouse_url(url: str) -> tuple[str, str, str, str]:
+    """Parse clickhouse://user:pass@host:port/db -> (http_url, db, user, password)."""
+    from urllib.parse import urlparse
+
+    parsed = urlparse(url.replace("clickhouse://", "http://"))
+    http_url = f"http://{parsed.hostname}:{parsed.port or 8123}"
+    db = (parsed.path or "/").strip("/") or "default"
+    user = parsed.username or "default"
+    password = parsed.password or ""
+    return http_url, db, user, password
 
 
 # ── Async helpers ────────────────────────────────────────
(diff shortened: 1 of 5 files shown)


Apoorvgarg-creator (commented):
## Code Review

Solid scaffolding and strong test layout, but several correctness and security issues must land before this is safe for production PII. Comments prefixed `bug` / `risky` / `nitpick` below.

**Must-fix before merge:** password in URL query string, missing TLS, SQL f-string injection via db identifier, FK case-mismatch false orphans, Replacing-idempotency gap, non-atomic streamed writes, and `raise` chains losing cause.

**Functionality gaps worth discussing:** no `--resume` for partial import failures, no `import_complete` marker in manifest, `export_time_cutoff` captured but not enforced in WHERE clause, no `--start-date` for incremental migrations.

observal_cli/cmd_migrate.py
```diff
@@ -330,6 +395,183 @@ async def _insert_table(
...
+    client: object | None = None,
+) -> object:
+    """Execute a ClickHouse query via HTTP.
+
+    If stream_to is provided, streams response body to disk.
+    An optional pre-existing client avoids creating new connections per call.
+    """
+    import httpx as _httpx
+
+    params = {"database": db, "user": user, "password": password}
```
Apoorvgarg-creator: `bug`: ClickHouse password transmitted as an HTTP query parameter lands in access logs, proxy logs, and `ps`-style process listings on intermediate hops. Query params are NOT confidential in HTTP. Use HTTP Basic Auth or the `X-ClickHouse-Key` header instead.

```python
# fix
resp = await client.post(
    http_url,
    content=sql,
    auth=(user, password),
    params={"database": db},
)
```
Why this matters: any operator with access to the ClickHouse server logs (or anything in between) can recover the DB password from routine logs.

observal_cli/cmd_migrate.py
```diff
@@ -186,6 +239,18 @@ def _sha256_file(path: Path) -> str:
     return h.hexdigest()
 
 
+def _parse_clickhouse_url(url: str) -> tuple[str, str, str, str]:
+    """Parse clickhouse://user:pass@host:port/db -> (http_url, db, user, password)."""
+    from urllib.parse import urlparse
+
+    parsed = urlparse(url.replace("clickhouse://", "http://"))
```
Apoorvgarg-creator: `bug`: hardcoded plaintext HTTP for a migration that carries PII (you even warn about PII at the end of `export_telemetry_cmd`). Credentials + telemetry cross the wire unencrypted. Support `clickhouses://` -> `https://` and allow port 8443. Also, `url.replace("clickhouse://", "http://")` is unanchored, so a password containing the literal scheme prefix corrupts parsing. Prefer:

```python
if url.startswith("clickhouse://"):
    url = "http://" + url[len("clickhouse://"):]
```
Without TLS this command is not safe to run outside a private network.

observal_cli/cmd_migrate.py
```diff
@@ -330,6 +395,183 @@ async def _insert_table(
...
+
+
+async def _ch_existing_tables(
+    http_url: str,
+    db: str,
+    user: str,
+    password: str,
+) -> set[str]:
+    """Query system.tables to discover which tables exist on target ClickHouse."""
+    sql = f"SELECT name FROM system.tables WHERE database = '{db}' FORMAT JSON"
```
Apoorvgarg-creator: `bug`: `db` is derived from a user-supplied URL and interpolated directly into SQL. A db name with a quote (`my'db`) breaks the query and opens injection. ClickHouse supports server-side parameters over HTTP:

```python
sql = "SELECT name FROM system.tables WHERE database = {db:String} FORMAT JSON"
# pass param_db=db alongside the query params
```
Audit every other `f"... {var} ..."` SQL in this file on the same principle. It's easy for identifiers to slip past review when the value looks "trusted" today.

observal_cli/cmd_migrate.py
```diff
@@ -618,6 +865,402 @@ async def _export_database(db_url: str, output_path: Path) -> ExportResult:
...
+
+    # Merge aliases
+    fk_values["mcp_id"] |= fk_values.pop("mcp_server_id", set())
+    fk_values["user_id"] |= fk_values.pop("actor_id", set())
+
+    # Filter to valid UUIDs only — ClickHouse stores these as String,
+    # so non-UUID values like "filesystem" or "default" can appear.
+    _uuid_re = re.compile(r"^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$", re.IGNORECASE)
+    for key in list(fk_values):
+        fk_values[key] = {v for v in fk_values[key] if _uuid_re.match(v)}
```
Apoorvgarg-creator: `bug`: PostgreSQL returns `uuid::text` canonicalized to lowercase, but ClickHouse stores UUIDs in `String` columns with arbitrary case. Your regex is case-insensitive so uppercase IDs survive the filter, but `ids - existing` is a case-sensitive set difference, so `"ABC..."` is flagged as orphaned even when the lowercase row exists in PG.

```python
fk_values[key] = {v.lower() for v in fk_values[key] if _uuid_re.match(v)}
```
Without this, validate-telemetry will report false orphans and block clean cutovers.

observal_cli/cmd_migrate.py
```diff
@@ -618,6 +865,402 @@ async def _export_database(db_url: str, output_path: Path) -> ExportResult:
...
+            rprint(f"  [yellow]Skipping {table_name} (table does not exist on target)[/yellow]")
+            tables_skipped.append(table_name)
+            rows_imported[table_name] = 0
+            continue
+
+        for filename in files:
+            filepath = input_dir / filename
+
+            # MergeTree idempotency: check if partition already has data
+            if table_cfg["engine"] == "mergetree":
```
Apoorvgarg-creator: `bug`: PR description advertises "idempotent re-import", but this gate runs only for MergeTree. ReplacingMergeTree re-inserts the rows and relies on background merges to de-dupe by sort key; until merge runs, the re-import doubles storage and duplicates are visible to any `SELECT` without `FINAL`. Apply the same partition-has-data check to replacing tables, or document that replacing tables require manual `TRUNCATE`/`ALTER TABLE ... DROP PARTITION` before re-import.

observal_cli/cmd_migrate.py
```diff
@@ -330,6 +395,183 @@ async def _insert_table(
...
+
+    params = {"database": db, "user": user, "password": password}
+    owns_client = client is None
+    if owns_client:
+        client = _httpx.AsyncClient(timeout=_httpx.Timeout(300.0, connect=10.0))
+    try:
+        if stream_to:
+            async with client.stream("POST", http_url, content=sql, params=params) as resp:
+                resp.raise_for_status()
+                with open(stream_to, "wb") as f:
```
Apoorvgarg-creator: `bug`: if the stream errors mid-write (network drop, ClickHouse closes connection), a half-written file stays at `stream_to`. Nothing cleans it up, and the checksum you compute right after will be for the *partial* file. Write to a tmp sibling and `os.replace()` only after the stream completes cleanly:

```python
tmp = stream_to.with_suffix(stream_to.suffix + ".tmp")
with open(tmp, "wb") as f:
    async for chunk in resp.aiter_bytes(chunk_size=65536):
        f.write(chunk)
os.replace(tmp, stream_to)
```
Atomic rename is the whole contract here — callers assume "file present" implies "file complete".

observal_cli/cmd_migrate.py
```diff
@@ -330,6 +395,183 @@ async def _insert_table(
...
+                        f.write(chunk)
+                return resp
+        else:
+            resp = await client.post(http_url, content=sql, params=params)
+            resp.raise_for_status()
+            return resp
+    except _httpx.HTTPStatusError as exc:
+        rprint(f"[red]ClickHouse returned HTTP {exc.response.status_code}[/red]")
+        rprint(f"[dim]{exc.response.text[:500]}[/dim]")
+        raise typer.Exit(1)
```
Apoorvgarg-creator: `bug`: `raise typer.Exit(1)` inside `except ... as exc:` without `from exc` drops the original traceback. When debugging you lose the real HTTP status / network error under a bare `typer.Exit`. Apply `from exc` (or `from None` if you explicitly want to suppress, which you don't here) on every re-raise — same issue at lines 440, 472, 475, 591, 612, 624, 682, 747, 884, 889, 905, 910, 1034, 1056, 1066, 1201, 1229, 1286, 1320, 1325, 1358, 1362, 1375, 1436, 1465, 1478. PEP 3134.

observal_cli/cmd_migrate.py
```diff
@@ -618,6 +865,402 @@ async def _export_database(db_url: str, output_path: Path) -> ExportResult:
...
+        raise
+    except Exception:
+        rprint("[red]ClickHouse health check failed.[/red]")
+        raise typer.Exit(1)
+
+    # Create output directory
+    if output_dir.exists() and any(output_dir.iterdir()):
+        rprint(f"[red]Output directory is not empty:[/red] {output_dir}")
+        raise typer.Exit(1)
+    os.makedirs(output_dir, mode=0o700, exist_ok=True)
```
Apoorvgarg-creator, line 982: `risky`: `os.makedirs(..., mode=0o700)` ANDs with process umask, so on umask 022 you get 0o755, not 0o700. Given the PII warning, this matters. Explicitly chmod after:

```python
Path(output_dir).mkdir(mode=0o700, parents=True, exist_ok=True)
os.chmod(output_dir, 0o700)
```
The umask behavior is a classic Unix gotcha — the mode argument is an upper bound, not a guarantee.

observal_cli/cmd_migrate.py
```diff
@@ -618,6 +865,402 @@ async def _export_database(db_url: str, output_path: Path) -> ExportResult:
...
+
+    dir_created = True
```
Apoorvgarg-creator: `risky`: `dir_created = True` is set unconditionally immediately before the `try:`, so inside `except Exception:` the guard `if dir_created ...` is always true. Any exception below this point triggers `rmtree` on whatever lives at `output_dir`. Either remove the flag or move the assignment so it only flips when *we* created the dir (versus the caller handing us one that already existed). Dead-flag destructive cleanup is exactly how users lose data.

observal_cli/cmd_migrate.py
```diff
@@ -618,6 +865,402 @@ async def _export_database(db_url: str, output_path: Path) -> ExportResult:
...
+        raise
+    except Exception:
+        rprint("[red]ClickHouse health check failed.[/red]")
+        raise typer.Exit(1)
+
+    existing = await _ch_existing_tables(http_url, db, user, password)
+    rows_imported: dict[str, int] = {}
+    tables_skipped: list[str] = []
+
+    for table_cfg in CLICKHOUSE_TABLES:
```
Apoorvgarg-creator, line 1172: `risky`: no atomicity across tables. If `spans` succeeds and `scores` fails mid-way, the target ClickHouse is left partial with no rollback hint. At minimum, write a `.import_state.json` file after each successful table so a re-run resumes cleanly, and surface the failed table clearly in CLI output. A multi-hour import that dies half-way with no resume is a painful operational experience.

observal_cli/cmd_migrate.py
```diff
@@ -25,6 +27,8 @@
 from observal_cli import client
 from observal_cli.render import spinner
 
+logging.getLogger("httpx").setLevel(logging.WARNING)
```
Apoorvgarg-creator: `risky`: mutating a global logger at module import time affects the whole process, including unrelated callers of `cmd_migrate` imports. Move this into the CLI command entrypoint (`main()` or the Typer callback) where logging config belongs. Import-time side effects make test setup and library reuse surprising.

observal_cli/cmd_migrate.py
```diff
@@ -330,6 +395,183 @@ async def _insert_table(
...
+
+async def _ch_query(
+    http_url: str,
+    db: str,
+    user: str,
+    password: str,
+    sql: str,
+    *,
+    stream_to: Path | None = None,
+    client: object | None = None,
```
Apoorvgarg-creator: `risky`: `client: object | None` defeats type-checking, and the name shadows the module-level `from observal_cli import client` for readers of this function. Rename to `http_client` and type it as `httpx.AsyncClient | None`; have `_ch_query` return `httpx.Response`. Also move `import httpx` to the module top since it's a declared dependency — in-function imports only make sense for optional deps or startup-latency-critical paths.

observal_cli/cmd_migrate.py
```diff
@@ -330,6 +395,183 @@ async def _insert_table(
...
+def _is_empty_parquet(path: Path) -> bool:
+    """Return True if the file is empty or a Parquet file with zero rows."""
+    if path.stat().st_size == 0:
+        return True
+    try:
+        import pyarrow.parquet as pq
+
+        meta = pq.read_metadata(path)
+        return meta.num_rows == 0
+    except Exception:
```
Apoorvgarg-creator: `risky`: bare `except Exception: return True` reports the file as empty for *any* pyarrow failure — corruption, permission errors, OOM — silently treating bad data as "no data". That is how rows disappear. Narrow to the pyarrow exceptions you actually expect (`pyarrow.lib.ArrowInvalid`, `pyarrow.lib.ArrowIOError`), log the rest with context, and let unexpected failures propagate.

observal_cli/cmd_migrate.py
```diff
@@ -594,6 +836,11 @@ async def _export_database(db_url: str, output_path: Path) -> ExportResult:
         # Ensure output parent directory exists
         output_path.parent.mkdir(parents=True, exist_ok=True)
 
+        # Write sidecar manifest for Phase 2 consumption
+        sidecar_stem = output_path.name.removesuffix(".tar.gz").removesuffix(".tgz")
+        sidecar_path = output_path.parent / f"{sidecar_stem}.manifest.json"
```
Apoorvgarg-creator, line 907: `risky`: the sidecar is an out-of-archive copy of `manifest.json`. If a user rotates/copies the archive but not the sidecar (or vice-versa), Phase 2 runs against a stale/mismatched manifest silently. Either (a) extract `manifest.json` from the tar.gz inside `_export_telemetry`, or (b) include a hash of the archive in the sidecar and verify it in Phase 2. Two sources of truth without a link between them always drift in production.

observal_cli/cmd_migrate.py
```diff
@@ -618,6 +865,402 @@ async def _export_database(db_url: str, output_path: Path) -> ExportResult:
...
+                        if val is not None and val != "":
+                            fk_values[col].add(str(val))
+
+    # Merge aliases
+    fk_values["mcp_id"] |= fk_values.pop("mcp_server_id", set())
+    fk_values["user_id"] |= fk_values.pop("actor_id", set())
+
+    # Filter to valid UUIDs only — ClickHouse stores these as String,
+    # so non-UUID values like "filesystem" or "default" can appear.
+    _uuid_re = re.compile(r"^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$", re.IGNORECASE)
```
Apoorvgarg-creator: `risky`: regex compiled on every call to `_validate_fk_references`. Hoist to module scope as a constant:

```python
_UUID_RE = re.compile(r"^[0-9a-f]{8}-...", re.IGNORECASE)
```
PEP 8 naming for module-level constants is UPPER_SNAKE. The cost of compilation is small, but this pattern repeated across a codebase adds up and reads as "regex just for this call".

observal_cli/cmd_migrate.py
```diff
@@ -90,6 +94,27 @@
     "exporter_configs": ["config"],
 }
 
+# ── Phase 2: ClickHouse telemetry constants ──────────────
+
+CLICKHOUSE_TABLES: list[dict[str, str | list[str]]] = [
```
Apoorvgarg-creator: `nitpick`: `list[dict[str, str | list[str]]]` forces every consumer to narrow with `cast` or accept `Any`. Use a `TypedDict`:

```python
from typing import Literal, TypedDict

class TableCfg(TypedDict):
    name: str
    engine: Literal["replacing", "mergetree"]
    time_col: str
    fk_cols: list[str]

CLICKHOUSE_TABLES: list[TableCfg] = [...]
```
mypy will catch typos in `table_cfg["engin"]` and flag impossible engine values — cheap win.

observal_cli/cmd_migrate.py
```diff
@@ -750,3 +1393,123 @@ def validate_cmd(
...
+
+
+# ── Phase 2: Telemetry CLI commands ─────────────────────
+
+
+@migrate_app.command("export-telemetry")
+def export_telemetry_cmd(
+    clickhouse_url: str = typer.Option(..., "--clickhouse-url", help="Source ClickHouse connection string"),
+    manifest: str = typer.Option(..., "--manifest", help="Path to Phase 1 migration_manifest.json"),
+    output_dir: str = typer.Option(..., "--output-dir", help="Directory for exported Parquet files"),
```
Apoorvgarg-creator, line 1518: `nitpick`: Typer natively supports `Path` with `exists/file_okay/dir_okay/resolve_path` checks. Replace `str` here and in the other telemetry commands with:

```python
output_dir: Path = typer.Option(
    ..., "--output-dir",
    file_okay=False, dir_okay=True, resolve_path=True,
)
```
Surfaces bad args at parse time, removes boilerplate `Path(...)` conversion, and matches how the rest of Typer-based CLIs usually look.

observal_cli/cmd_migrate.py
```diff
@@ -618,6 +865,402 @@ async def _export_database(db_url: str, output_path: Path) -> ExportResult:
...
+                "orphaned_agent_ids_truncated": False,
+                "orphaned_mcp_ids": [],
+                "orphaned_mcp_ids_truncated": False,
+                "orphaned_user_ids": [],
+                "orphaned_user_ids_truncated": False,
+                "validated_at": None,
+            },
+        }
+        manifest_out = output_dir / "telemetry_manifest.json"
+        manifest_out.write_text(json.dumps(telemetry_manifest, indent=2) + "\n")
```
Apoorvgarg-creator: `nitpick`: `write_text` uses the platform default encoding. On Windows that's cp1252, which mangles non-ASCII JSON. Always pin `encoding="utf-8"` on text I/O, and do the same for every `read_text`/`write_text` in this PR — JSON manifests must be UTF-8 per RFC 8259.

observal_cli/cmd_migrate.py
```diff
@@ -618,6 +865,402 @@ async def _export_database(db_url: str, output_path: Path) -> ExportResult:
...
+        raise typer.Exit(1)
+    p1_manifest = json.loads(manifest_path.read_text())
+    if not p1_manifest.get("phase1_completed_at"):
+        rprint("[red]Phase 1 has not completed.[/red]")
+        rprint("[dim]  Run 'observal migrate export' and 'observal migrate import' first.[/dim]")
+        raise typer.Exit(1)
+    migration_id = p1_manifest["migration_id"]
+
+    # Record cutoff before any queries
+    export_time_cutoff = datetime.now(UTC).isoformat()
```
Apoorvgarg-creator: `nitpick`: you capture the cutoff *before* queries (good) but never use it as an upper bound in `_build_ch_*_query`. The point of a cutoff is `WHERE time_col < cutoff` so late writes during export don't get partially included. Either enforce it in the WHERE clause or drop the field — a manifest field called "cutoff" that isn't enforced is actively misleading for anyone reading the manifest later.

observal_cli/cmd_migrate.py
```diff
@@ -618,6 +865,402 @@ async def _export_database(db_url: str, output_path: Path) -> ExportResult:
...
+        rprint(f"  [green]✓[/green] {table_name}: {rows_imported[table_name]:,} rows")
+
+    elapsed = time.monotonic() - t0
+    return TelemetryImportResult(
+        migration_id=migration_id,
+        tables_imported=len([t for t in rows_imported if rows_imported[t] > 0]),
+        tables_skipped=tables_skipped,
+        rows_imported=rows_imported,
+        duration_seconds=round(elapsed, 2),
+        warnings=warnings,
```
Apoorvgarg-creator, line 1227: `nitpick`: `warnings` is initialized and passed through but never appended to anywhere. Either populate it (skipped tables, partition-skipped files, slow imports) so operators have a single place to look for non-fatal issues, or remove the field from `TelemetryImportResult`. Empty plumbing rots — it reads as "someone planned to add this" forever.

observal_cli/cmd_migrate.py
```diff
@@ -618,6 +865,402 @@ async def _export_database(db_url: str, output_path: Path) -> ExportResult:
...
+            rprint(f"  Importing {filename}...")
+            await _ch_import(http_url, db, user, password, table_name, filepath)
+
+        rows_imported[table_name] = table_info.get("row_count", 0)
+        rprint(f"  [green]✓[/green] {table_name}: {rows_imported[table_name]:,} rows")
+
+    elapsed = time.monotonic() - t0
+    return TelemetryImportResult(
+        migration_id=migration_id,
+        tables_imported=len([t for t in rows_imported if rows_imported[t] > 0]),
```
Apoorvgarg-creator: `nitpick`: materializes an intermediate list just to count. Use a generator:

```python
tables_imported=sum(1 for v in rows_imported.values() if v > 0),
```
No correctness change, just tighter and more idiomatic.

Apoorvgarg-creator (approved):
LGTM

PR #461, opened 2026-04-20
fix: correct agent name resolution and 404 error hint

### LINKED ISSUE #457: CLI suggests incorrect command `observal registry agent list` — should be `observal agent list`
## Description

After certain CLI operations, the suggestion text tells the user to run `observal registry agent list`, but this command doesn't exist. The correct command is `observal agent list`.

## Current Behavior

The CLI outputs a suggestion like:

```
Run `observal registry agent list` to see your agents.
```

Running this command fails because it's not a valid subcommand.

## Expected Behavior

The suggestion should display the correct command:

```
Run `observal agent list` to see your agents.
```

## Proposed Solution

Update the suggestion/hint string in the CLI to reference `observal agent list` instead of `observal registry agent list`.

### LINKED ISSUE #458: `observal pull` fails on first run but succeeds after `observal agent list`
## Description

Running `observal pull <agent> --ide kiro` fails with a "Not found" error on the first attempt, but succeeds if you run `observal agent list` beforehand. This suggests some local cache or state initialization is required before `pull` works, but the user isn't told this.

## Steps to Reproduce

1. `observal pull test123 --ide kiro`
2. Error: `Not found (/api/v1/agents/test123). Check that the resource ID is correct, or use observal registry agent list to browse.`
3. Run `observal agent list`
4. Run `observal pull test123 --ide kiro` again — succeeds

## Expected Behavior

`observal pull` should work on the first attempt without requiring a prior `agent list` call. If there's a cache/sync step needed, `pull` should handle it internally.

## Proposed Solution

Investigate why `agent list` unblocks `pull`. Likely the list command populates a local cache or refreshes an auth token that `pull` depends on. The `pull` command should either:
- Perform the same initialization internally before attempting the fetch, or
- Provide a clear error message explaining the prerequisite step

Image

Image

ShaanNarendran:
## Purpose / Description
Two related bugs around agent name handling in the CLI and server:

1. **Wrong CLI hint on 404**: The 404 error handler in `client.py` always suggested `observal registry agent list`, but `agent` is a top-level subcommand — the correct command is `observal agent list`.
2. **`observal pull` fails on first run by name** (#458): `_load_agent` only tried name-based lookup for agents created by the current user. When referencing another user's agent by name (e.g. `observal pull appian_docs_v1 --ide kiro`), the name fallback missed it and returned 404. Running `observal agent list` first worked around this by populating the local name→UUID cache.

## Fixes
* Fixes #458
* Fixes #457 

## Approach
- **`observal_cli/client.py`**: In the 404 handler, check if the resource type is `agent` and suggest `observal agent list` instead of `observal registry agent list`.
- **`observal-server/api/routes/agent.py`**: Add a global name fallback in `_load_agent` after the user-specific lookup. When exactly one agent matches the name, return it regardless of creator. When multiple agents share the same name, return `None` to avoid ambiguity.

## How Has This Been Tested?

- Verified the 404 error path produces the correct hint for both agent and registry resource types
- Confirmed `_load_agent` resolves agent names globally when the user-specific lookup misses

The `resolve_alias` function in the CLI resolves names via a local cache populated by `list` commands. When the cache is empty (first run), the raw name is passed to the server. The server's `_load_agent` was only doing a user-scoped name fallback, which silently failed for agents created by other users.

## Checklist

### DIFF
--- a/observal-server/api/routes/agent.py
+++ b/observal-server/api/routes/agent.py
@@ -75,6 +75,14 @@ async def _load_agent(
             if mine:
                 return mine
 
+        # Fall back to global name lookup (any creator)
+        stmt = select(Agent).where(Agent.name == agent_id).options(*_agent_load_options)
+        if extra_conditions:
+            stmt = stmt.where(*extra_conditions)
+        results = (await db.execute(stmt)).scalars().all()
+        if len(results) == 1:
+            return results[0]
+
         return None
 
 
--- a/observal-server/api/routes/agent.py
+++ b/observal-server/api/routes/agent.py
@@ -47,12 +47,14 @@ async def _load_agent(
     extra_conditions=None,
     *,
     prefer_user_id: uuid.UUID | None = None,
+    org_id: uuid.UUID | None = None,
 ) -> Agent | None:
     """Load an agent by UUID, prefix, or name with eager loading.
 
     When *prefer_user_id* is provided and resolution is by name, prefer the
     caller's own agent over agents created by other users with the same name.
-    Returns None if no match is found for the current user.
+    The global name fallback is restricted to active agents and, when *org_id*
+    is set, to agents within the same organisation.
     """
     try:
         return await resolve_prefix_id(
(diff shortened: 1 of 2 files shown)


Haz3-jolt:
Needs some changes for org_id for multi tenancy stuff, I am pushing

PR #462, opened 2026-04-20
feat: redesign traces list page and add session filtering, user resolution

Kaushik-Kumar-CEG:
## Summary

- Add traces list page with 8-column table: session, user, platform, tokens, tools, duration, started, score
- Resolve user display names from PostgreSQL for all sessions including OTLP-hash-only and Kiro sessions
- Support platform and time range server-side filters, client-side search, and active/all tab toggle
- Add /sessions/summary endpoint for daily session count
- Propagate username through all hook scripts via OBSERVAL_USERNAME env var and X-Observal-Username header
- Store user.name from hook payloads in ClickHouse during ingestion
- Display green left-edge bar for active sessions, colored token counts for Claude Code, orange credits for Kiro

## NOTE 

there's still some issues with kiro hooks - regarding the credits and stuff
also the username schema doesnt work well with kiro for some reason
issues will be filed and touched up soon

image

## Test plan

- [x] Traces page loads sessions on first visit without refresh
- [x] User column shows display names, not raw hashes
- [x] Platform and time range dropdowns filter correctly
- [x] Active sessions show green left-edge indicator
- [x] Claude Code sessions show input/output token counts
- [x] Kiro sessions show credit usage
- [x] ruff, tsc, and next lint pass with no errors

### DIFF
--- a/observal-server/api/routes/otel_dashboard.py
+++ b/observal-server/api/routes/otel_dashboard.py
@@ -3,13 +3,16 @@ import json
 import logging
 import re
 import time as _time
+import uuid as _uuid
 from datetime import UTC, datetime
 
 from fastapi import APIRouter, Depends, Query, Request
 from fastapi_cache.decorator import cache
+from sqlalchemy import select
 
 from api.deps import require_role
 from config import settings
+from database import async_session
 from models.user import User, UserRole
 from services.clickhouse import _query, query_shim_spans_for_window
 from services.redis import publish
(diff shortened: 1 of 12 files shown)


Haz3-jolt:
Pushed `6a08a82` with review fixes:

**Platform filter pushed to ClickHouse** - The platform filter was being applied in Python after `LIMIT 100`, which meant requesting `platform=kiro` could return far fewer than 100 results even when more existed. Moved it into a `HAVING` clause with `multiIf` normalization so legacy service names (`observal-hooks`, `kiro-cli`, `observal-shim`) are normalized in the SQL itself before filtering.

**Days parameter capped at 365** - Previously unbounded, so `?days=999999` would scan the entire ClickHouse table. Now capped before it hits the query.

**Targeted user resolution** - The old code did `SELECT id, email, name FROM users` (full table scan) whenever any session lacked a `user_name`. Now it collects only the specific unresolved `user_id` values, partitions them into UUIDs vs emails, and queries with `WHERE id IN (...) OR email IN (...)`.

**Removed sessions/summary cache** - The `@cache` decorator's key builder only uses path + query params, not the current user. Since admins see all sessions and regular users see only their own, an admin's cached response could be served to a non-admin (or vice versa). Removed the 15s cache since it's a lightweight aggregation anyway.

**OTEL_RESOURCE_ATTRIBUTES merge** - The reconciler was unconditionally overwriting this env var, silently dropping any user-defined resource attributes. Now it parses existing comma-separated pairs and merges Observal's `user.id=...` into them, preserving everything else.

**Collapsed redundant if branches** - Lines 148-151 had two separate `if` statements (`if not row.get("user_name") and not is_admin` / `if not row.get("user_name") and is_admin`) that both assigned `current_user.name`. Since one of `is_admin`/`not is_admin` is always true, these collapsed into a single `if not row.get("user_name")`.

**Removed redundant `import json as _json`** in `kiro_stop_hook.py`, since `json` is already imported at module level.

**Added trust-model comment** on the unauthenticated hooks endpoint noting that `user_name` and `user_id` are self-reported, and that the `/sessions` endpoint resolves canonical names from PostgreSQL as the authoritative source.

PR #482, opened 2026-04-21
fix(hooks): resolve server URL from auth config instead of hardcoding localhost

Haz3-jolt:
## Purpose / Description
Hook scripts were hardcoded to fall back to `http://localhost:8000` when no `--url` flag or `OBSERVAL_HOOKS_URL` env var was set. This meant hooks silently pointed to localhost regardless of what server the user authenticated against during `observal auth login`.

## Fixes
* Hooks now resolve the server URL from `~/.observal/config.json` (the `server_url` saved during `observal auth login`) before falling back to localhost.

## Approach
Added a unified URL resolution order across all 5 hook scripts and the server-side install endpoint:

1. `--url` CLI flag (Kiro Python hooks only)
2. `OBSERVAL_HOOKS_URL` env var
3. `~/.observal/config.json` → `server_url` field
4. `http://localhost:8000` as last resort

**Files changed:**
- **`kiro_hook.py`** / **`kiro_stop_hook.py`** — Added `_resolve_hooks_url()` that checks env var and config file before falling back. `--url` flag still takes priority.
- **`flush_buffer.py`** — Same `_resolve_hooks_url()` instead of bare `os.environ.get(..., "localhost:8000")`.
- **`observal-hook.sh`** / **`observal-stop-hook.sh`** — Added config file lookup via inline Python when `OBSERVAL_HOOKS_URL` isn't set.
- **`hook.py` (server-side)** — `install_hook` route now derives `server_url` from the incoming `Request` object instead of using the hardcoded default in `generate_hook_telemetry_config`.

## How Has This Been Tested?
- `make test` — 98 tests in `test_agent_config_generator.py` pass
- Verified URL resolution order: `--url` > env var > config file > localhost fallback

## Checklist

### DIFF
--- a/observal-server/api/routes/agent.py
+++ b/observal-server/api/routes/agent.py
@@ -709,9 +709,11 @@ async def install_agent(
     # Resolve all component names for rules file content
     name_map = await _resolve_component_names(agent.components, db)
 
+    server_url = str(request.base_url).rstrip("/")
     snippet = generate_agent_config(
         agent,
         req.ide,
+        observal_url=server_url,
         mcp_listings=mcp_listings_map,
         component_names=name_map,
         env_values=req.env_values,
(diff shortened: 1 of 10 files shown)


Haz3-jolt:
Superseded by #484 which implements full server-side endpoint discovery.

PR #495, opened 2026-04-22
feat: Kimi eval provider + dashboard scores + ClickHouse init hardening

swathisaravananan:
## Summary
- **Eval:** adds a Moonshot/Kimi judge provider. Setting `EVAL_MODEL_PROVIDER=moonshot` (or a model name containing `kimi`) routes through the OpenAI-compatible path with Moonshot's base URL and `thinking: {type: disabled}` so Kimi returns a single deterministic JSON score instead of streaming reasoning tokens. Moonshot's Instant Mode requires `temperature=0.6`, which is set automatically.
- **Dashboard:** displays per-dimension and overall agent display scores.
- **Docker:** hardens first-boot ClickHouse init — the healthcheck now passes `CLICKHOUSE_PASSWORD`, and `entrypoint.sh` creates the target DB over HTTP with exponential-backoff retries before running `init_clickhouse()`, so races against the clickhouse container no longer wedge startup.

## Verification
- `pytest tests/eval/test_eval_phase8.py` — 19/19 pass, incl. two new `TestMoonshotProvider` tests covering both explicit-provider and auto-detect-by-model-name paths.
- Live smoke-test against `kimi-k2.5`: all 8 `EVAL_TEMPLATES` returned well-formed `{score, reason}` payloads with coherent scoring (e.g. `tool_selection_accuracy=1.0`, `graph_context_precision=0.5` when 1 of 2 retrieved chunks was off-topic, `recall_accuracy=0.3` for tangentially relevant memory).

## Test plan
- [ ] `pytest tests/eval/test_eval_phase8.py`
- [ ] Set `EVAL_MODEL_PROVIDER=moonshot`, `EVAL_MODEL_NAME=kimi-k2.5`, `EVAL_MODEL_API_KEY=...` and run an eval on a recent trace; confirm scores land in ClickHouse and render on the dashboard.
- [ ] Fresh `docker compose up` on a machine without prior ClickHouse state; confirm server comes up without manual intervention and healthcheck passes with a non-default `CLICKHOUSE_PASSWORD`.

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 set -e
 
 echo "Ensuring base schema exists..."
@@ -19,43 +19,39 @@ echo "Running database migrations..."
 /app/.venv/bin/python -m alembic upgrade head
 
 echo "Ensuring ClickHouse database exists..."
-# Parse credentials from CLICKHOUSE_URL (clickhouse://user:pass@host:port/db)
-CH_URL="${CLICKHOUSE_URL:-clickhouse://default:clickhouse@observal-clickhouse:8123/observal}"
-CH_HOST=$(echo "$CH_URL" | sed 's|clickhouse://[^@]*@||' | sed 's|/.*||')
-CH_USER=$(echo "$CH_URL" | sed 's|clickhouse://||' | sed 's|:.*||')
-CH_PASS=$(echo "$CH_URL" | sed 's|clickhouse://[^:]*:||' | sed 's|@.*||')
-CH_DB=$(echo "$CH_URL" | sed 's|.*/||')
+# Parse credentials from CLICKHOUSE_URL using Python to handle special chars
+eval "$(/app/.venv/bin/python -c "
+from urllib.parse import urlparse, unquote
+import os
+url = urlparse(os.environ.get('CLICKHOUSE_URL', 'clickhouse://default:clickhouse@observal-clickhouse:8123/observal'))
+print(f'CH_HOST={url.hostname}:{url.port or 8123}')
+print(f'CH_USER={unquote(url.username or \"default\")}')
+print(f'CH_PASS={unquote(url.password or \"clickhouse\")}')
+print(f'CH_DB={url.path.lstrip(\"/\") or \"observal\"}')
+")"
 CH_PROTO_HOST="http://${CH_HOST}"
 
 CH_CREATE_RETRIES=15
 CH_CREATE_COUNT=0
 while [ $CH_CREATE_COUNT -lt $CH_CREATE_RETRIES ]; do
-  HTTP_CODE=$(/app/.venv/bin/python -c "
-import urllib.request, sys
-try:
-    req = urllib.request.Request('${CH_PROTO_HOST}/?user=${CH_USER}&password=${CH_PASS}',
-                                data=b'CREATE DATABASE IF NOT EXISTS ${CH_DB}')
-    urllib.request.urlopen(req, timeout=5)
-    print('ok')
-except Exception:
-    print('fail')
-")
+  HTTP_CODE=$(wget -q -O- --user="$CH_USER" --password="$CH_PASS" \
+    --post-data="CREATE DATABASE IF NOT EXISTS ${CH_DB}" \
+    "${CH_PROTO_HOST}/" 2>/dev/null && echo "ok" || echo "fail")
   if [ "$HTTP_CODE" = "ok" ]; then
     echo "ClickHouse database '${CH_DB}' ready"
     break
   fi
   CH_CREATE_COUNT=$((CH_CREATE_COUNT + 1))
-  WAIT=$((2 ** CH_CREATE_COUNT > 30 ? 30 : 2 ** CH_CREATE_COUNT))
-  echo "Waiting for ClickHouse to accept connections (attempt $CH_CREATE_COUNT/$CH_CREATE_RETRIES, retry in ${WAIT}s)..."
-  sleep "$WAIT"
   if [ $CH_CREATE_COUNT -ge $CH_CREATE_RETRIES ]; then
     echo "ClickHouse not reachable after $CH_CREATE_RETRIES attempts, aborting."
     exit 1
   fi
+  WAIT=$(( 2 ** CH_CREATE_COUNT > 30 ? 30 : 2 ** CH_CREATE_COUNT ))
+  echo "Waiting for ClickHouse to accept connections (attempt $CH_CREATE_COUNT/$CH_CREATE_RETRIES, retry in ${WAIT}s)..."
+  sleep "$WAIT"
 done
 
 echo "Initializing ClickHouse tables..."
-# Retry logic for ClickHouse initialization
 MAX_RETRIES=10
 RETRY_COUNT=0
 while [ $RETRY_COUNT -lt $MAX_RETRIES ]; do
@@ -70,9 +66,9 @@ asyncio.run(init_clickhouse())
   else
     RETRY_COUNT=$((RETRY_COUNT + 1))
     if [ $RETRY_COUNT -lt $MAX_RETRIES ]; then
-      WAIT_TIME=$((2 ** RETRY_COUNT))
+      WAIT_TIME=$(( 2 ** RETRY_COUNT > 30 ? 30 : 2 ** RETRY_COUNT ))
       echo "ClickHouse initialization failed. Retrying in ${WAIT_TIME}s (attempt $RETRY_COUNT/$MAX_RETRIES)..."
-      sleep $WAIT_TIME
+      sleep "$WAIT_TIME"
     else
       echo "ClickHouse initialization failed after $MAX_RETRIES attempts"
       exit 1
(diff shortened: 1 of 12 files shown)


docker/entrypoint.sh
```diff
@@ -18,12 +18,58 @@ asyncio.run(init())
 echo "Running database migrations..."
 /app/.venv/bin/python -m alembic upgrade head
 
+echo "Ensuring ClickHouse database exists..."
+# Parse credentials from CLICKHOUSE_URL (clickhouse://user:pass@host:port/db)
+CH_URL="${CLICKHOUSE_URL:-clickhouse://default:clickhouse@observal-clickhouse:8123/observal}"
+CH_HOST=$(echo "$CH_URL" | sed 's|clickhouse://[^@]*@||' | sed 's|/.*||')
+CH_USER=$(echo "$CH_URL" | sed 's|clickhouse://||' | sed 's|:.*||')
+CH_PASS=$(echo "$CH_URL" | sed 's|clickhouse://[^:]*:||' | sed 's|@.*||')
```
Apoorvgarg-creator: risky: password parsing here breaks if the password contains `:` or `@`, and it isn't URL-decoded either. Since we interpolate `${CH_PASS}` straight into the query string a few lines down, a password with `&`, `#`, `+`, or `%` will either corrupt the query or silently auth-fail. Can we switch to `curl -u "$CH_USER:$CH_PASS"` (or `wget --user/--password`) and let the tool handle encoding? That also sidesteps the `sed` brittleness entirely.

docker/entrypoint.sh
```diff
@@ -18,12 +18,58 @@ asyncio.run(init())
...
+CH_DB=$(echo "$CH_URL" | sed 's|.*/||')
+CH_PROTO_HOST="http://${CH_HOST}"
+
+CH_CREATE_RETRIES=15
+CH_CREATE_COUNT=0
+while [ $CH_CREATE_COUNT -lt $CH_CREATE_RETRIES ]; do
+  HTTP_CODE=$(wget -q -O- "${CH_PROTO_HOST}/?user=${CH_USER}&password=${CH_PASS}" \
```
Apoorvgarg-creator: risky: the password ends up in the URL query string here, which means it's visible in `ps` inside the container and in any proxy/access logs we ever put in front of CH. ClickHouse HTTP supports basic auth and the `X-ClickHouse-Key` header. Could we move auth off the query string before this merges? Happy to take it as a follow-up if you'd rather keep the diff small, but I'd want a TODO at minimum.

docker/entrypoint.sh
```diff
@@ -18,12 +18,58 @@ asyncio.run(init())
...
+    --post-data="CREATE DATABASE IF NOT EXISTS ${CH_DB}" 2>/dev/null && echo "ok" || echo "fail")
+  if [ "$HTTP_CODE" = "ok" ]; then
+    echo "ClickHouse database '${CH_DB}' ready"
+    break
+  fi
+  CH_CREATE_COUNT=$((CH_CREATE_COUNT + 1))
+  WAIT=$((2 ** CH_CREATE_COUNT > 30 ? 30 : 2 ** CH_CREATE_COUNT))
```
Apoorvgarg-creator: risky: shebang is `#!/bin/sh` but we're using `**` for exponentiation. POSIX `sh` doesn't specify `**`, and on some base images `/bin/sh` resolves to dash or busybox `ash`, where this either errors out or silently evaluates to `0` (which would make the sleep a no-op and hammer the CH container in a tight loop). Do you know which shell this actually runs under in our image? If bash is guaranteed, I'd just flip the shebang to `#!/bin/bash`. Otherwise we can replace the `**` with a small multiplication loop.

docker/entrypoint.sh
```diff
@@ -18,12 +18,58 @@ asyncio.run(init())
...
+  echo "Waiting for ClickHouse to accept connections (attempt $CH_CREATE_COUNT/$CH_CREATE_RETRIES, retry in ${WAIT}s)..."
+  sleep "$WAIT"
+  if [ $CH_CREATE_COUNT -ge $CH_CREATE_RETRIES ]; then
```
Apoorvgarg-creator, line 45: nitpick: this `-ge CH_CREATE_RETRIES` check runs after the `sleep` on the previous line, so on the final failing attempt we wait up to 30s before aborting. Small thing, but flipping the check above the sleep (or just relying on the outer `while` condition and printing the abort message after the loop) gives a faster fail and removes the redundancy with the loop condition.

docker/entrypoint.sh
```diff
@@ -18,12 +18,58 @@ asyncio.run(init())
...
 
 asyncio.run(init_clickhouse())
-"
+"; then
+    echo "ClickHouse initialization successful"
+    break
+  else
+    RETRY_COUNT=$((RETRY_COUNT + 1))
+    if [ $RETRY_COUNT -lt $MAX_RETRIES ]; then
+      WAIT_TIME=$((2 ** RETRY_COUNT))
```
Apoorvgarg-creator: risky: this backoff isn't capped, so attempt 10 sleeps `2^10 = 1024s` (~17 min) before giving up, and the total worst case is close to 34 min. That's well past what most orchestrators (and humans running `docker compose up`) will tolerate before assuming it's wedged. Can we cap this the same way as the create loop above (`WAIT_TIME=$((2 ** RETRY_COUNT > 30 ? 30 : 2 ** RETRY_COUNT))`) or drop `MAX_RETRIES` to something like 6?

web/src/app/(registry)/agents/[id]/page.tsx
```diff
@@ -211,22 +211,41 @@ function AnalyticsTab({ agentId }: { agentId: string }) {
...
+                thought_process: "Thought Process",
+              };
+              const dimColors: Record<string, string> = {
+                goal_completion: "bg-emerald-500",
+                tool_efficiency: "bg-blue-500",
+                tool_failures: "bg-amber-500",
+                factual_grounding: "bg-violet-500",
+                thought_process: "bg-cyan-500",
+              };
+              const clampedScore = Math.max(0, Math.min(100, Number(score)));
```
Comment (Apoorvgarg-creator) [inline web/src/app/(registry)/agents/[id]/page.tsx:218]: doubt: I want to make sure I'm reading this right. The old code was `Math.min(Number(score) * 10, 100)`, which implied the backend was returning 0 to 10 here. The new code drops the `* 10` and treats `score` as 0 to 100 (and we render `/100` next to it). Did the aggregate endpoint start returning 0 to 100, or are we reading the same 0 to 10 field and silently collapsing the bars to under 10% width? A quick curl snippet against the endpoint, or a pointer to where the scale changed, would help me sign off on this one. If the scale did flip, worth adding a small render test so we don't regress it silently.

web/src/app/(admin)/dashboard/page.tsx
```diff
@@ -61,6 +62,35 @@ function TopDownloadsBar({ items }: { items: { name: string; value: number }[] }
...
+        <div className="flex items-center justify-between">
+          <span className="font-[family-name:var(--font-display)] text-sm font-semibold truncate">
+            {agent.name}
+          </span>
+          {latest.version && (
+            <Badge variant="secondary" className="text-[10px]">v{latest.version}</Badge>
+          )}
+        </div>
+        <ScoreOverview
+          displayScore={latest.display_score ?? latest.overall_score ?? 0}
```
Apoorvgarg-creator: doubt: `latest.display_score ?? latest.overall_score ?? 0` worries me a little. `display_score` is 0 to 10 (we render it with `/10`), but `overall_score` has historically looked like a 0 to 1 value in other surfaces. If a record ever has `overall_score` without `display_score`, we'd render something like `0.8/10`, which looks broken rather than wrong. Is the fallback actually load-bearing for any current record, or can we drop it and show a placeholder when `display_score` is null? Same question applies on the eval detail page where you wired the same fallback.

Apoorvgarg-creator (commented):
Added some minor comments

observal-server/services/eval/eval_service.py
```diff
@@ -101,6 +101,8 @@ async def call_eval_model(prompt: str) -> dict:
 
     if provider == "bedrock" or (not provider and "anthropic" in eval_model):
         return await _call_bedrock(prompt, eval_model)
+    if provider == "moonshot" or (not provider and "kimi" in eval_model.lower()):
+        return await _call_openai_compatible(prompt, eval_model, provider="moonshot")
```
Apoorvgarg-creator, line 105: doubt: this picked up the same Moonshot branch as `eval_engine.py`, but I don't see a matching test under `tests/eval/` for this call site. Is one covering the other via a shared path I'm missing? If not, I'd prefer either mirroring the two new `TestMoonshotProvider` tests here, or (cleaner) extracting a `_build_openai_body(provider, model, prompt)` helper so both `_call_openai` and `_call_openai_compatible` share the Moonshot-specific knobs in one place. Right now the Kimi defaults live in two files that can drift.

web/src/components/dashboard/score-overview.tsx
```diff
@@ -0,0 +1,160 @@
...
+
+export function ScoreOverview({
+  displayScore,
+  grade,
+  dimensionScores,
+  penaltyCount,
+  compact = false,
+}: ScoreOverviewProps) {
+  const dims = Object.entries(dimensionScores).filter(
+    ([key, value]) => value !== null && key !== "adversarial_robustness"
```
Apoorvgarg-creator: nitpick: `key !== "adversarial_robustness"` is doing a load-bearing filter that isn't obvious from the call sites. Could we express this next to `DIMENSION_META` (e.g. a `hiddenInOverview: true` flag on the entry) so future readers can see which dimensions are excluded without grepping the component body? Not blocking, just saves a future debugging session when someone adds a new dimension and wonders why it's missing from the card.

PR #515, opened 2026-04-22
fix: restart nginx LB after local `make rebuild`

ShaanNarendran:
## Purpose / Description
After `make rebuild` / `make reset` / `make rebuild-clean`, the API container gets a new IP but nginx keeps routing to the old cached one, causing 502 errors locally.

PR #503 fixed this for CI deploys but the local Makefile targets were missed.

## Fixes
* Complements #503 — extends the nginx restart fix to local dev workflows

## Approach
Added `docker compose restart observal-lb` after the API health check in the `rebuild`, `reset`, and `rebuild-clean` Makefile targets, matching the same approach used in the CI deploy workflow.

## How Has This Been Tested?

- Ran `make rebuild` and verified `curl localhost:8000/health` returns 200 immediately after (previously returned 502)

## Checklist


Haz3-jolt (approved):
LGTM

PR #535, opened 2026-04-23
fix(hooks): skip trailing metadata in stop hook transcript parsing

Haz3-jolt:
## Purpose / Description

The stop hook was never capturing assistant responses or thinking blocks from Claude Code sessions.

## Fixes
* Fixes assistant response and thinking traces never appearing in Observal

## Approach

The transcript JSONL ends with metadata/attachment entries *after* the last user prompt. Reading backwards with `tac`, the hook hit a `user` type entry before reaching any `assistant` entries, so it always broke with zero captures.

Fix: only break on user messages after we've found at least one assistant message. This skips past the trailing metadata to reach the actual assistant turn.

**Before:** `MSG=0 THINK=0` (always empty)
**After:** `MSG=1 THINK=0` (captures response; thinking when present)

## How Has This Been Tested?

Tested against a live 32MB transcript (15K lines) from a real Claude Code session. Verified the loop now correctly captures the most recent assistant text response.

## Checklist

### DIFF
--- a/observal_cli/hooks/observal-stop-hook.sh
+++ b/observal_cli/hooks/observal-stop-hook.sh
@@ -42,12 +42,16 @@ trap 'rm -rf "$TMPDIR_WORK"' EXIT
 
 MSG_COUNT=0
 THINK_COUNT=0
+FOUND_ASSISTANT=0
 
 # Use process substitution instead of pipe to avoid subshell variable scoping.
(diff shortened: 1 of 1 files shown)


Haz3-jolt:
Already merged to main as c33b7c4. Agent attribution work moved to #536.

PR #590, opened 2026-04-25
fix: prevent duplicate Observal hooks in Kiro agent configs

### LINKED ISSUE #589: Duplicate traces in Kiro integration: hooks accumulate on every agentSpawn
## Bug

Every Kiro hook event fires multiple times (5-7x) because `_auto_inject_hooks()` in `observal_cli/hooks/kiro_hook.py` appends Observal hooks to agent configs without deduplicating.

### Root Cause

1. The guard check on line 116 has a no-op duplicate `or` condition (`"telemetry/hooks" in cmd or "telemetry/hooks" in cmd`) — same expression on both sides.
2. Even when the guard works, `merged.setdefault(evt, []).extend(entries)` blindly appends, so duplicates accumulate permanently.
3. Three different sources produce hooks with different python paths (`python3`, `sys.executable`, `cat | python3 /full/path/...`), so the guard doesn't always match.
4. `cmd_doctor.py`'s `_install_kiro_hooks` has a similar append-without-strip pattern.

### Impact

- Duplicate events in ClickHouse per user action
- Inflated prompt/tool/token counts in the traces UI
- Redundant traces appearing in the frontend

### Expected Behavior

Exactly 1 Observal hook per event type per agent config, regardless of how many times `_auto_inject_hooks`, `observal doctor`, or `observal pull` runs.

### Affected Files

- `observal_cli/hooks/kiro_hook.py` — `_auto_inject_hooks()`
- `observal_cli/cmd_doctor.py` — `_install_kiro_hooks()`

HemalathaMadeswaran18:
Strip existing Observal hooks before adding the canonical one in both _auto_inject_hooks (kiro_hook.py) and _install_kiro_hooks (cmd_doctor.py).

## Purpose / Description
Every Kiro hook event fires 5-7x because _auto_inject_hooks() in kiro_hook.py
  accumulates duplicate Observal hooks in agent configs without deduplicating.
  Three different code paths (observal pull, observal doctor/observal auth
  login, and _auto_inject_hooks on every agentSpawn) each append hooks with
  slightly different python paths (python3, sys.executable, cat | python3
  /full/path/...), and the guard meant to prevent re-injection has a no-op
  duplicate or condition. This results in inflated prompt/tool/token counts and
  redundant traces in the frontend UI.

## Fixes
* Fixes #589 

## Approach
 Added an _is_observal_hook(h) helper that identifies any hook belonging to Observal by
  checking for observal_cli or telemetry/hooks in the command string — this catches all three
  command formats regardless of python path.
  
  Changed both injection sites to strip-then-add instead of check-then-append:
  
  - kiro_hook.py _auto_inject_hooks(): Removed the broken guard + .extend(). Now filters out
  all existing Observal hooks per event, then appends exactly one canonical hook.
  - cmd_doctor.py _install_kiro_hooks(): Same pattern — strip existing Observal hooks before
  adding the desired ones, instead of skipping the whole event if any Observal hook is
  detected.
  
  This is idempotent: running it N times always results in exactly 1 Observal hook per event
  type.

## How Has This Been Tested?

  - All 296 tests in test_ide_config_e2e.py and test_agent_config_generator.py pass
  - All 42 kiro/hook-specific tests pass (pytest -k "kiro or hook or inject")
  - Manual testing: simulated 3 separate Kiro chat sessions (different session IDs) with
  cooldown stamp cleared between each to force re-injection — verified exactly 1 Observal hook
  per event after all 3 runs
  - Verified _is_observal_hook correctly matches all 3 command formats found in the wild
  (python3 -m, sys.executable -m, cat | python3 /full/path/...)

  - The root cause was a copy-paste typo: "telemetry/hooks" in cmd or "telemetry/hooks" in cmd
   — identical expressions on both sides of or, making it a no-op duplicate check
  - Even with a correct guard, a check-then-skip approach is fragile when multiple code paths
  write to the same config file with different command formats. Strip-then-add is more robust
  since it's idempotent by design
  - No server-side or frontend dedup exists in the pipeline (ClickHouse ingestion, session
  queries, or React UI), so preventing duplicates at the source is the only fix

## Checklist

### DIFF
--- a/observal_cli/cmd_doctor.py
+++ b/observal_cli/cmd_doctor.py
@@ -1144,15 +1144,15 @@ def _install_kiro_hooks(server_url: str) -> tuple[list[str], bool]:
         current_hooks = data.get("hooks", {})
         updated = False
 
+        def _is_observal_hook(h: dict) -> bool:
+            cmd = h.get("command", "")
+            return "observal_cli" in cmd or "telemetry/hooks" in cmd
+
         for kiro_event, desired_entries in desired_kiro_hooks.items():
             existing = current_hooks.get(kiro_event, [])
-            # Check if Observal hook already present
-            has_observal = any(
-                "observal" in h.get("command", "") or "telemetry/hooks" in h.get("command", "") for h in existing
-            )
-            if not has_observal:
-                # Append our hooks, keep existing ones
-                current_hooks[kiro_event] = existing + desired_entries
+            cleaned = [h for h in existing if not _is_observal_hook(h)]
+            current_hooks[kiro_event] = cleaned + desired_entries
+            if current_hooks[kiro_event] != existing:
                 updated = True
 
         if updated:
--- a/observal_cli/hooks/kiro_hook.py
+++ b/observal_cli/hooks/kiro_hook.py
@@ -94,10 +94,17 @@ def _maybe_auto_inject(url: str):
         pass
 
 
+def _is_observal_hook(h: dict) -> bool:
+    """Return True if a hook entry belongs to Observal."""
+    cmd = h.get("command", "")
+    return "observal_cli" in cmd or "telemetry/hooks" in cmd
+
+
 def _auto_inject_hooks(url: str):
     """Inject Observal hooks into any Kiro agent configs that lack them.
 
     Runs only on agentSpawn events so new agents get hooks on first use.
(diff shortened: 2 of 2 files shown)


HemalathaMadeswaran18:
testing evidence;
image

PR #591, opened 2026-04-26
feat: add Prometheus metrics endpoint and alerting rules

### LINKED ISSUE #212: Prometheus exporter - metrics endpoint and alerting rules
## Problem

Observal has no `/metrics` endpoint for Prometheus scraping. The only monitoring surface is the `/health` endpoint, which returns a simple JSON status. Operators cannot integrate Observal with Prometheus-based alerting or Grafana dashboards without manual ClickHouse queries. There are no pre-built alerting rules for critical conditions like ingestion failures, high error rates, or ClickHouse latency spikes.

## Goal

Expose a Prometheus-compatible `/metrics` endpoint on the FastAPI server and ship a set of default alerting rules, giving operators production-grade observability out of the box.

## Scope (checklist)

- [ ] Add `prometheus-fastapi-instrumentator` to `pyproject.toml` dependencies
- [ ] Instrument the FastAPI `app` in `main.py` (auto HTTP request latency, status codes, in-progress requests)
- [ ] Add custom application metrics in `services/clickhouse.py` (ClickHouse query duration histogram, query error counter)
- [ ] Add custom business metrics in `api/routes/telemetry.py` and `api/routes/otlp.py` (traces ingested counter, spans ingested counter, ingestion error counter, token usage counter)
- [ ] Expose active session gauge and per-harness request labels
- [ ] Create `prometheus/alerting-rules.yml` with rules: high error rate (>5% 5m), ingestion stall (0 spans 10m), ClickHouse query p99 >2s, health endpoint down
- [ ] Create `prometheus/grafana-dashboard.json` for the Prometheus datasource (request rate, latency percentiles, ingestion throughput, token usage)
- [ ] Add Prometheus service to `docker/docker-compose.yml` scraping `observal-api:8000/metrics`
- [ ] Add `METRICS_ENABLED` toggle to `config.py` Settings

## Affected Files

- `observal-server/pyproject.toml`
- `observal-server/main.py`
- `observal-server/config.py`
- `observal-server/services/clickhouse.py`
- `observal-server/api/routes/telemetry.py`
- `observal-server/api/routes/otlp.py`
- `docker/docker-compose.yml`
- `prometheus/alerting-rules.yml` (new)
- `prometheus/grafana-dashboard.json` (new)

DEVAANSH001:
## Purpose / Description

Observal had no `/metrics` endpoint for Prometheus scraping. The only monitoring surface was the `/health` endpoint returning a simple JSON status. Operators couldn't integrate Observal with Prometheus-based alerting or Grafana dashboards without running manual ClickHouse queries, and there were no pre-built alerting rules for critical conditions like ingestion failures, high error rates, or ClickHouse latency spikes.

## Fixes

* Fixes #212

## Approach

Added a Prometheus-compatible `/metrics` endpoint to the FastAPI server and shipped default alerting rules and a Grafana dashboard for production-grade observability out of the box.

**Instrumentation (`main.py`, `config.py`)**
- Added `METRICS_ENABLED: bool = True` toggle to `Settings` so operators can disable the endpoint via environment variable
- Mounted `prometheus-fastapi-instrumentator` after middleware setup — auto-instruments all HTTP routes with request count, latency histograms, and in-progress gauges

**ClickHouse metrics (`services/clickhouse.py`)**
- Added `clickhouse_query_duration_seconds` histogram with `operation` label instrumented around `_query()`
- Added `clickhouse_query_errors_total` counter incremented on every caught exception before re-raising

**Business metrics (`api/routes/telemetry.py`, `api/routes/otlp.py`)**
- Added `observal_traces_ingested_total` and `observal_spans_ingested_total` counters
- Added `observal_ingestion_errors_total` with `type` label
- Added `observal_token_usage_total` with `type` label (`input`/`output`)

**Alerting (`prometheus/alerting-rules.yml`)**

9 rules covering all 4 required by the issue plus 5 bonus rules:

| Severity | Rule | Condition |
|---|---|---|
| Critical | `HealthEndpointDown` | API unreachable 2m |
| Critical | `IngestionStall` | 0 spans ingested 10m |
| Critical | `ClickHouseConnectionErrors` | Query errors >1/sec 5m |
| Warning | `HighErrorRate` | 4xx/5xx rate >5% over 5m |
| Warning | `ClickHouseSlowQueries` | p99 latency >2s over 5m |
| Warning | `HighIngestionErrorRate` | Ingestion errors >10% over 5m |
| Warning | `LowIngestionThroughput` | Rate <20% of baseline 15m |
| Warning | `HighAPILatency` | p95 latency >5s over 5m |
| Warning | `HighMemoryUsage` | Memory >90% of limit 5m |

**Infrastructure (`docker/docker-compose.yml`, `prometheus/prometheus.yml`)**
- Added `observal-prometheus` service scraping `observal-api:8000/metrics` every 15s
- Auto-provisioned Grafana Prometheus datasource via `grafana/provisioning/datasources/prometheus.yaml`
- Pre-built 6-panel Grafana dashboard at `grafana/dashboards/prometheus-metrics.json`

## How Has This Been Tested?

Verified full stack end-to-end with `docker compose up -d` (18 services).

**Generated traffic to populate dashboard panels:**

```powershell
# Normal traffic
for ($i=1; $i -le 50; $i++) {
    Invoke-WebRequest -Uri "http://localhost:8000/health" -UseBasicParsing | Out-Null
    Invoke-WebRequest -Uri "http://localhost:8000/metrics" -UseBasicParsing | Out-Null
    Start-Sleep -Milliseconds 200
}

# Error traffic — intentional 4xx requests
for ($i=1; $i -le 20; $i++) {
    try { Invoke-WebRequest -Uri "http://localhost:8000/nonexistent" -UseBasicParsing | Out-Null } catch {}
    try { Invoke-WebRequest -Uri "http://localhost:8000/api/fake" -UseBasicParsing | Out-Null } catch {}
    Start-Sleep -Milliseconds 100
}

# Ingestion test
observal ops telemetry test  # Output: Test event sent! Ingested: 1
```

**Results verified:**
- `http://localhost:8000/metrics` returns Prometheus text format with all custom metrics
- Prometheus target `observal-api` shows **State: UP** scraping every 15s
- All 9 alerting rules loaded from `/etc/prometheus/alerting-rules.yml` — all **INACTIVE** (system healthy)
- Grafana dashboard auto-provisioned under Observal folder with all 6 panels showing live data
- Error Rate gauge showing **30.9%** from intentional error traffic
- `METRICS_ENABLED=false` returns `404`, re-enabling returns `200`

> **Note:** Active session gauge and per-IDE labels were not implemented in this PR 
> as the current codebase has no session tracking mechanism to hook into. 
> This can be added in a follow-up once session management is in place.

## Checklist

- [x] You have a descriptive commit message with a short title (first line, max 50 chars)
- [x] UI changes: screenshots of all affected screens included below

---

### SS1 - All 18 services healthy including `observal-prometheus`
Screenshot 2026-04-26 010349

### SS2 - `/metrics` Endpoint Live
Screenshot 2026-04-26 011423

### SS3 - Prometheus Targets UP
Screenshot 2026-04-26 012324

### SS4 — All 9 Alerting Rules 
Screenshot 2026-04-26 013110

### SS5 - Prometheus Graph with Real Data
Screenshot 2026-04-26 013435

### SS6 - Grafana Datasources
Screenshot 2026-04-26 132326

### SS7 - Grafana Dashboard: All 6 Panels with Live Data 
Screenshot 2026-04-26 120352


observal-server/api/routes/telemetry.py
```diff
@@ -736,6 +770,8 @@ async def ingest(
         except Exception:
             logger.exception("Failed to insert scores")
             errors += len(batch.scores)
+            ingestion_errors.labels(type="scores").inc(len(batch.scores))
```
Apoorvgarg-creator, line 791: **bug**: This line duplicates the `errors += len(batch.scores)` that is already two lines above (line 771 in the PR). You will end up double-counting score insertion failures in the `IngestResponse`. Looks like the prometheus counter line was inserted between the two identical statements during editing.

Just remove this second one:

```python
        except Exception:
            logger.exception("Failed to insert scores")
            errors += len(batch.scores)
            ingestion_errors.labels(type="scores").inc(len(batch.scores))
            # delete the line below -- it's a duplicate
            errors += len(batch.scores)
```

grafana/provisioning/datasources/prometheus.yaml
```diff
@@ -3,8 +3,11 @@ apiVersion: 1
 datasources:
   - name: Prometheus
     type: prometheus
-    uid: observal-prometheus
+    uid: prometheus
```
Apoorvgarg-creator: **risky**: You are renaming the datasource UID from `observal-prometheus` to `prometheus`. Every existing Grafana dashboard in this repo (there are 10 of them in `grafana/dashboards/`) that references the old UID will break -- they will show "datasource not found" until manually re-pointed.

The new dashboard you are adding (`prometheus-metrics.json`) uses `"uid": "prometheus"`, but all the existing dashboards were authored against `observal-prometheus`.

Two options:
1. Keep the existing UID (`observal-prometheus`) and update your new dashboard JSON to use it instead.
2. If you really want to change the UID, do a find-and-replace across all existing dashboard JSON files so nothing breaks.

Option 1 is simpler and safer. No reason to rename an internal UID.

docker/docker-compose.yml
```diff
@@ -266,12 +249,59 @@ services:
     networks:
       - observal-net
 
+  observal-prometheus:
```
Apoorvgarg-creator, line 214: **risky**: You are downgrading the Prometheus image from `v3.4.2` (what is currently on `main`) to `v3.1.0`. The comment says "Stable LTS release, tested in production" but Prometheus does not have a formal LTS program -- v3.4.2 is actually more recent and already proven in this project.

Unless you hit a specific bug in v3.4.2, please keep the existing version. Downgrades can cause TSDB format incompatibilities and data loss if someone already has a `prometheusdata` volume created by the newer version.

Suggestion:
```yaml
    image: prom/prometheus:v3.4.2
```

observal-server/services/clickhouse.py
```diff
@@ -61,19 +75,34 @@ async def _query(sql: str, params: dict | None = None, *, data: str | None = Non
...
-        "password": CLICKHOUSE_PASSWORD,
-    }
-    # Inject admin-configured resource overrides (e.g. max_memory_usage)
-    if _resource_overrides:
-        query_params.update(_resource_overrides)
-    if params:
-        query_params.update(params)
-    body = f"{sql}\n{data}" if data else sql
-    return await client.post(CLICKHOUSE_HTTP, content=body, params=query_params)
+    start_time = time.time()
```
Apoorvgarg-creator: **risky**: This restructure wraps the entire `_query()` body in a try/except. The function is the lowest-level ClickHouse call -- every single read and write goes through it. A few concerns:

1. The `operation` detection (`"INSERT" in sql.upper()`) will incorrectly label `SELECT ... INSERT` subqueries or `DESCRIBE` / `ALTER` / `CREATE TABLE` statements as "select". For DDL that runs during `init_clickhouse()`, the label is misleading. Consider adding an `operation` parameter or at least handling "alter"/"create" as a third bucket.

2. On success you record duration, but on exception you don't. If a query times out after 10 seconds you lose that latency data point -- which is arguably the most interesting one. Move the `observe()` call into a `finally` block so you always get it.

3. The `time.time()` call is fine but `time.monotonic()` is more correct for measuring elapsed durations since it is not affected by system clock adjustments.

Suggested structure:
```python
async def _query(sql, params=None, *, data=None):
    start = time.monotonic()
    operation = "insert" if "INSERT" in sql.upper() else "select"
    try:
        # ... existing logic ...
        return response
    except Exception:
        clickhouse_query_errors.labels(operation=operation).inc()
        raise
    finally:
        duration = time.monotonic() - start
        clickhouse_query_duration.labels(operation=operation).observe(duration)
```

observal-server/main.py
```diff
@@ -240,6 +240,13 @@ async def dispatch(self, request: Request, call_next):
 # --- GZip compression for responses >= 500 bytes ---
 app.add_middleware(GZipMiddleware, minimum_size=500)
 
+# --- Prometheus instrumentation ---
+if settings.METRICS_ENABLED:
+    # Expose /metrics only when monitoring is enabled.
```
Apoorvgarg-creator: **doubt**: The instrumentator is now placed right after `GZipMiddleware` but before `CacheControlMiddleware` and all the route registrations. In the original code on `main`, it was placed after all routes were registered (line 299-302).

`prometheus-fastapi-instrumentator` needs to see registered routes to generate per-handler labels. If you call `.instrument(app)` before routes are registered via `include_router()`, the auto-discovered handler list may be incomplete and you could get metrics bucketed under a generic label instead of per-endpoint.

Can you verify that the `/api/v1/telemetry/ingest`, `/api/v1/telemetry/hooks`, and other routes actually show up with their handler names in the `/metrics` output? If not, move the instrumentator block back to after the `include_router()` calls but keep the `METRICS_ENABLED` guard:

```python
app.include_router(config_router)

if settings.METRICS_ENABLED:
    Instrumentator(
        excluded_handlers=["/livez", "/healthz", "/readyz", "/metrics"],
    ).instrument(app).expose(app, endpoint="/metrics", include_in_schema=False)
```

observal-server/api/routes/telemetry.py
```diff
@@ -38,6 +39,27 @@
 
 logger = logging.getLogger(__name__)
 router = APIRouter(prefix="/api/v1/telemetry", tags=["telemetry"])
+
+# Prometheus metrics for telemetry ingestion
+traces_ingested = Counter(
```
Apoorvgarg-creator: **nitpick**: These metric objects are module-level globals that get created at import time regardless of the `METRICS_ENABLED` setting. When `METRICS_ENABLED=false`, the `/metrics` endpoint returns 404 (which is correct), but the counters still get incremented in memory on every request. It won't cause errors but it is wasted work.

Not blocking, but if you want to be consistent with the toggle, you could wrap these in a lazy initialization or check the setting. Something for a follow-up.

observal-server/api/routes/telemetry.py
```diff
@@ -551,15 +573,24 @@ async def ingest(
                 )
             await insert_traces(rows)
             ingested += len(rows)
+            # Count successfully persisted trace rows.
+            traces_ingested.inc(len(rows))
         except Exception:
             logger.exception("Failed to insert traces")
```
Apoorvgarg-creator, line 598: **nitpick**: You are incrementing `traces_ingested` after `insert_traces()` succeeds, which is the right place. But the counter is called `observal_traces_ingested_total` while what you are actually counting here is trace *rows* prepared for insertion. If a batch has 5 traces, you increment by 5 -- that is fine, just calling it out so the naming is intentional.

Also, small thing -- the token tracking loop inside the spans block (lines 582-586 in the diff) runs before `insert_spans()`. If the insert fails, you have already counted those tokens but they were never persisted. Consider moving the token counting after the successful `insert_spans()` call, same pattern you used for `spans_ingested`.

prometheus/alerting-rules.yml
```diff
@@ -0,0 +1,133 @@
...
+          severity: warning
+          component: api
+        annotations:
+          summary: "High error rate detected"
+          description: "Error rate is {{ $value | humanizePercentage }} (threshold: 5%)"
+
+      # Ingestion stall alert
+      - alert: IngestionStall
+        expr: |
+          rate(observal_spans_ingested_total[10m]) == 0
```
Apoorvgarg-creator, line 25: **nitpick**: The `IngestionStall` rule uses `rate(observal_spans_ingested_total[1h] offset 1h) > 0` as a guard to avoid false alarms on fresh deployments. That is a smart idea. However, `rate()` over a 1h window with a 1h offset means you need at least 2 hours of metric history before this rule can even evaluate to true. Might be worth mentioning in a comment or docs that this alert will not fire for the first couple hours after initial deployment.

prometheus/alerting-rules.yml
```diff
@@ -0,0 +1,133 @@
...
+          severity: warning
+          component: api
+        annotations:
+          summary: "High API request latency"
+          description: "P95 latency for {{ $labels.handler }} is {{ $value }}s (threshold: 5s)"
+
+      # Memory pressure (requires node_exporter)
+      - alert: HighMemoryUsage
+        expr: |
+          (
```
Apoorvgarg-creator, line 123: **doubt**: The `HighMemoryUsage` alert uses `container_memory_usage_bytes` and `container_spec_memory_limit_bytes` -- these come from cAdvisor, not from the Observal API's `/metrics` endpoint. The current `prometheus.yml` only scrapes `observal-api:8000`, so Prometheus will never receive these metrics and this rule will never fire.

Either:
1. Add a cAdvisor or node-exporter scrape target to `prometheus.yml`
2. Or mark this rule clearly as "requires additional scrape config" so operators don't wonder why it is always inactive

Right now it will just sit there silently doing nothing, which could give false confidence.

grafana/dashboards/prometheus-metrics.json
```diff
@@ -0,0 +1,136 @@
...
+  "panels": [
+    {
+      "id": 1,
+      "title": "HTTP Request Rate",
+      "type": "timeseries",
+      "datasource": {"type": "prometheus", "uid": "prometheus"},
+      "gridPos": {"h": 8, "w": 12, "x": 0, "y": 0},
+      "targets": [
+        {
+          "datasource": {"type": "prometheus", "uid": "prometheus"},
```
Apoorvgarg-creator: **doubt**: The dashboard queries reference `http_requests_total` and `http_request_duration_seconds_bucket`, but `prometheus-fastapi-instrumentator` by default exposes metrics named `http_request_duration_seconds` (histogram), `http_requests_total` (counter), and `http_requests_in_progress` (gauge). The exact names depend on the instrumentator version and config.

Can you confirm these metric names match what you actually see in the `/metrics` output? If the instrumentator uses `http_request_duration_highres_seconds` instead (which some versions do), these panels will show "No data".

Also, the Error Rate gauge on line 38 filters on `status="4xx"` but Prometheus labels from the instrumentator typically use the full status code (e.g., `status="404"`) not the class. You would need a regex match like `status=~"4.."` or `status=~"5.."`. Double-check the actual label values.

docker/docker-compose.yml
```diff
@@ -266,12 +249,59 @@ services:
...
+    image: prom/prometheus:v3.1.0  # Stable LTS release, tested in production
+    ports:
+      - "${PROMETHEUS_HOST_PORT:-9090}:9090"
+    # Explicit command flags for data persistence and web console
+    # --storage.tsdb.path: Critical for data persistence across container restarts
+    # --config.file: Explicit config location for clarity
+    # --web.console.*: Enables built-in Prometheus web console for debugging
+    command:
+      - '--config.file=/etc/prometheus/prometheus.yml'
+      - '--storage.tsdb.path=/prometheus'
```
Apoorvgarg-creator, line 220: **nitpick**: The inline comments here are quite heavy for a docker-compose file. Comments like "# Config file: Scrape configuration for observal-api metrics endpoint" and "# Data volume: Persists metrics data across container restarts (critical)" describe what is already self-evident from the YAML keys.

The project's existing docker-compose style (look at the other services above) uses minimal or no inline comments. I would trim these down to just the non-obvious ones -- the `command` flags explanation is fine to keep since those flags are less discoverable, but the volume and healthcheck comments can go.

docker/docker-compose.yml
```diff
@@ -266,12 +249,59 @@ services:
...
+        condition: service_healthy
+    restart: unless-stopped
+    # Health check: Ensures Prometheus is ready before Grafana connects
+    # Prevents connection failures during startup
+    healthcheck:
+      test: ["CMD", "wget", "--spider", "-q", "http://localhost:9090/-/healthy"]
+      interval: 10s
+      timeout: 5s
+      retries: 3
+    # Security: Prevent privilege escalation (production best practice)
```
Apoorvgarg-creator: **nitpick**: The memory breakdown comment is oddly precise ("Base Prometheus: ~150MB, Metric storage: ~50MB, Alert rules: ~20MB, Query cache: ~100MB, Headroom: ~192MB"). This level of detail will be wrong the moment the metric cardinality changes, and someone might trust it more than they should. 512M is a reasonable limit -- you don't need to justify it with made-up numbers. A simple "512M handles typical metric volumes" is enough.

prometheus/prometheus.yml
```diff
@@ -0,0 +1,28 @@
+global:
+  scrape_interval: 15s
+  evaluation_interval: 15s
+  external_labels:
+    cluster: 'observal'
```
Apoorvgarg-creator, line 5: **nitpick**: `external_labels` with `environment: 'production'` is hardcoded. For a self-hosted tool that many people will run in dev/staging, this will be misleading in any multi-cluster Prometheus federation setup. Consider making this configurable or just removing it since it is only meaningful in federated setups that this default config does not target.

observal-server/config.py
```diff
@@ -109,6 +109,9 @@ def validate_retention_days(cls, v: int) -> int:
     DEMO_USER_EMAIL: str | None = None
     DEMO_USER_PASSWORD: str | None = None
 
+    # Prometheus metrics
```
Apoorvgarg-creator: **nitpick**: Minor placement thing -- the `METRICS_ENABLED` setting is sandwiched between the demo account settings and `model_config`. It would read more naturally grouped with the other infrastructure settings (near `REDIS_URL`, `CLICKHOUSE_URL`, etc.) or at least near the `DEPLOYMENT_MODE` block. Not a big deal but makes it easier to find.

Apoorvgarg-creator (commented):
Please take a look at the minor comments

grafana/dashboards/prometheus-metrics.json
DEVAANSH001: Confirmed from live `/metrics` output during testing
 
These Metric names are present:
- `http_requests_total`
- `http_request_duration_seconds_bucket`
- `http_request_duration_highr_seconds_bucket` (high-res variant, also present but not used in dashboard)

Status label values - the instrumentator uses bucketed strings not raw status codes:`
 
So `status="4xx"` is correct  `status=~"4.."` would actually return no data since the label value is the string `"4xx"` not `"404"`. I checked that during testing before building the dashboard panels.

Image

Image
 
Image

Image

docker/docker-compose.yml
DEVAANSH001: In my last pr remark i was told to document well that's why,  i will remove it no problem.

observal-server/main.py
DEVAANSH001: Yes you are right so i have moved the Prometheus instrumentator block to after all router registrations so it can see the full route set before exposing `/metrics` with `METRICS_ENABLED` guard.

prometheus/alerting-rules.yml
DEVAANSH001: Good points. I've removed the `HighMemoryUsage` alert since it requires cAdvisor metrics which are not currently scraped in this base config. I also added a note about the 2-hour history requirement for the `IngestionStall` alert to prevent confusion on new deployments.

DEVAANSH001:
@Apoorvgarg-creator - Thanks so much for the detailed review and taking the time to go through everything so thoroughly. Your feedback was really helpful and I've addressed all the points you raised. Really appreciate it! If something doesn't fit properly, please let me know.

DEVAANSH001:
How should i fix this now ? 
image
this was the conflict part

ShaanNarendran:
@Apoorvgarg-creator

ShaanNarendran:
@DEVAANSH001 could you rebase it without conflicts and add tests?

DEVAANSH001:
Let me create a new pr this is a mess now or is this fixable ? @ShaanNarendran

ShaanNarendran:
> Let me create a new pr this is a mess now or is this fixable ? @ShaanNarendran

Yep we will merge that much quicker, sorry for the hold up on this @DEVAANSH001

PR #644, opened 2026-04-29
fixed several issues in kiro session delineation and visible frontend…

### LINKED ISSUE #642: Seperate chat instances showing up in same chat on observal UI - kiro-classic
### Checked for duplicates?

create an agent on observal, then open terminal
1. kiro-cli --classic --agent <agentname>
2. have a dummy chat
3. ctrl+c
4. repeat several times
5. check observal ui, thise chats show up in the same chat.

### Expected behaviour

**What Happened**
Although several different chat instances were opened on kiro-cli classic, they all got logged under one chat in observal UI
**What should have happened**
the chats should have showed up seperately

While applying a fix, please ensure you verify that all other attributes like model, credits etc get populated correctly. 

**Cause**
Reuse of pid in kiro-scli, there we need to either explore a different approach to solve this issue.

VishnuM049:
… spans

## Purpose / Description
Different chat instances in kiro-cli show up as one session in the frontend, and chat sessions created in different terminal windows also go to the same session. 
This happens because the current method uses PPID, which changes for every hook, and these are grouped by a backend cache that merges all events from the same directory within a 30 minute window into one session. 
Due to the above issue, Claude Code MCP initialization spans showed up in Kiro sessions due to the way clickhouse.py captures pre-hook spans.

## Fixes
* Fixes #642

## Approach
Captures the kiro-cli PID once via a shell environment variable (KIRO_CLI_PID=$PPID) and uses it to build a session ID (kiro-cli-{PID}). Every hook in the same run gets the same ID, restarting produces a new one. This ensures the session is tied to the actual kiro-cli process, not hooks or the backend cache. Avoids the backend cache by failing it's regex check.

## How Has This Been Tested?
-Started kiro-cli, stopped, restarted in same terminal and same wroking directory
-Tested with new terminal windows and concurrent sessions as well
-Queried database directly to verify distinct session IDs per run
-Confirmed MCP spans from other IDEs no longer leak into Kiro traces

_Describe the research stage_

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,77 @@
 
 All notable changes to this project will be documented in this file.
 
+## [0.4.0] - 2026-04-28
+
+### Added
+
+- add account page with profile, theme picker, and notifications (**web**) ([d748558](https://github.com/BlazeUp-AI/Observal/commit/d7485587c3bbdfd043fdfe58eb6bc4c4170758e1))
+- add 10 curated IDE themes with hand-tuned oklch palettes (**web**) ([dcbb704](https://github.com/BlazeUp-AI/Observal/commit/dcbb7047bb31d3fc0fd590b9092b5bbe3270e8ce))
+- add `observal doctor cleanup` command (**cli**) ([ab26241](https://github.com/BlazeUp-AI/Observal/commit/ab26241ab2a129a9479e92b78425138dc5477e8e))
+
+### Changed
+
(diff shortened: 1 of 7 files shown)


Haz3-jolt (commented):
~130 lines of code (`_get_parent_pid`, `_is_kiro_cli_process`, `_find_kiro_cli_pid`) are duplicated verbatim between `kiro_hook.py` and `kiro_stop_hook.py`. If a bug gets fixed in one copy it's easy to miss the other.

Worth extracting into a shared module like `observal_cli/hooks/_kiro_utils.py` and importing from both files. The `PROCESSENTRY32` struct is also defined twice within each file (once in each function) so that could be a single module-level definition too.

Logic itself looks correct, no functional issues spotted.

PR #648, opened 2026-04-30
feat(server): implement agent registry RBAC

### LINKED ISSUE #209: [SEC-AUTH] Team-level agent access control
## Problem

Agents support only two visibility modes: public or single-user owned. Authorization checks are limited to `created_by == current_user.id`. There is no way for multiple team members to collaboratively manage an agent. The `Organization` model is flat with no membership table.

## Goal

Introduce team-level access control so agents can be owned and permissioned by a team within an organization.

## Scope (checklist)

- [ ] Add `Team` and `TeamMembership` models (user_id, team_id, role: viewer/editor/admin)
- [ ] Add `AgentTeamAccess` model (agent_id, team_id, permission: view/edit/publish)
- [ ] Add `AgentVisibility` enum (`public`, `org_private`, `team_private`) to Agent
- [ ] Create `effective_agent_permission()` dependency in `api/deps.py`
- [ ] Update agent CRUD routes to enforce team permissions
- [ ] Update `list_agents` query to filter by visibility + team membership
- [ ] Add team management endpoints (create, invite, set role)
- [ ] Add Alembic migration for new tables
- [ ] Extend to other resources (MCP listings, prompts, hooks, skills) as follow-up

## Affected Files

- `observal-server/models/organization.py`
- `observal-server/models/agent.py`
- `observal-server/schemas/agent.py`
- `observal-server/api/deps.py`
- `observal-server/api/routes/agent.py`
- New: `observal-server/api/routes/team.py`
- `observal-server/models/__init__.py`

harishankar0301:
## Purpose / Description
Introduces enterprise-grade team-level access control and visibility settings for Agents. This replaces the previous binary public/single-owner model with a flexible visibility system that supports organizational collaboration via private agents and granular team-based permissions.

## Fixes
* Fixes #209

## Approach
* **Database & Schema:** Replaced `is_private` boolean with `visibility` (`AgentVisibility` enum: `public`/`private`). Added an `agent_team_access` table linked via foreign key with CASCADE deletes to prevent orphaned records.
* **Authentication & JWT:** Updated the OAuth flow and token generation to extract SSO `groups` and embed them as an ephemeral claim within the JWT payload, avoiding unnecessary database writes.
* **Authorization System:** Built a new `get_effective_agent_permission` evaluator that computes a 4-tier hierarchy (`owner` > `edit` > `view` > `none`) based on creator status, admin roles, and JWT SSO group overlap.
* **API Endpoints:** Modified all read/write endpoints for Agents to enforce RBAC rules, returning 403 Forbidden for unauthorized access while safely embedding computed `user_permission` statuses in successful responses. Configured a full-replacement strategy for updating team permissions.
* **Frontend UI:** Created the `AccessSettingsWidget` for the Agent detail page and added a "Visibility & Access" config section to the Agent Builder. Included a `useUpdateAgent` TanStack query mutation for optimistic saves.
* **Enterprise Gating:** Leveraged `DEPLOYMENT_MODE == "enterprise"` to selectively hide features from the UI in cloud or self-hosted environments.

## How Has This Been Tested?
Executed `make test` from the root of the project to verify schema migrations, API permission logic, mock data accuracy,

## UI Changes

image

<p> Agent visibility control for each agent based on groups</p>
image


Haz3-jolt:
Fixed the lint failure (unsorted imports in the alembic migration) and corrected the DCO sign-off format on the first commit. Also removed debug `print(payload)` in `deps.py` and `print(token)` in `auth.py` that would leak JWT payloads and OAuth tokens to stdout in production.

A few things to address before merge:

The `GET /agents/{id}/traces` endpoint doesn't call `get_effective_agent_permission()`, so private agents' trace data is accessible to anyone who knows the agent ID. Add the same check pattern used on the other endpoints.

`TeamAccessRequest.group_name` is a bare `str` with no constraints. Add `min_length=1, max_length=255` to prevent empty strings or excessively long values.

The migration drops several unrelated indexes (`ix_hook_listings_bundle_id`, `ix_mcp_listings_bundle_id`, etc.) and has a `DROP TABLE IF EXISTS` that looks like autogenerated Alembic noise. Strip unrelated changes from the migration - dropping those indexes could degrade query performance.

Groups array is stored in both access and refresh tokens. Orgs with many AD groups per user could bloat token size significantly. Consider capping the number of groups or storing them server-side.

No dedicated tests for `get_effective_agent_permission()` logic (group membership, no groups + private, admin override, etc.). Would be good to add a focused test class for the permission evaluator.

Haz3-jolt:
> Fixed the lint failure (unsorted imports in the alembic migration) and corrected the DCO sign-off format on the first commit. Also removed debug `print(payload)` in `deps.py` and `print(token)` in `auth.py` that would leak JWT payloads and OAuth tokens to stdout in production.
> 
> A few things to address before merge:
> 
> The `GET /agents/{id}/traces` endpoint doesn't call `get_effective_agent_permission()`, so private agents' trace data is accessible to anyone who knows the agent ID. Add the same check pattern used on the other endpoints.
> 
> `TeamAccessRequest.group_name` is a bare `str` with no constraints. Add `min_length=1, max_length=255` to prevent empty strings or excessively long values.
> 
> The migration drops several unrelated indexes (`ix_hook_listings_bundle_id`, `ix_mcp_listings_bundle_id`, etc.) and has a `DROP TABLE IF EXISTS` that looks like autogenerated Alembic noise. Strip unrelated changes from the migration - dropping those indexes could degrade query performance.
> 
> Groups array is stored in both access and refresh tokens. Orgs with many AD groups per user could bloat token size significantly. Consider capping the number of groups or storing them server-side.
> 
> No dedicated tests for `get_effective_agent_permission()` logic (group membership, no groups + private, admin override, etc.). Would be good to add a focused test class for the permission evaluator.

I have the bandwidth so I am addressing it myself.

harishankar0301:
Instead of including all directory groups for a user, we can configure the IAM provider to emit only the subset of groups that are relevant to the application (i.e., groups assigned to Observal). This effectively limits the token to the intersection of:

* Groups the user is a member of
* Groups explicitly assigned to the application

By scoping group claims this way, we can significantly reduce token size and avoid unnecessary bloat.

This would be a change on the IAM side.

Haz3-jolt:
> Instead of including all directory groups for a user, we can configure the IAM provider to emit only the subset of groups that are relevant to the application (i.e., groups assigned to Observal). This effectively limits the token to the intersection of:
> 
>     * Groups the user is a member of
> 
>     * Groups explicitly assigned to the application
> 
> 
> By scoping group claims this way, we can significantly reduce token size and avoid unnecessary bloat.
> 
> This would be a change on the IAM side.

Done, pushed into your PR. Can you give a glance through before we merge?

Haz3-jolt (approved):

PR #649, opened 2026-04-30
feat(cli): add observal support bundle and inspect commands

naraen-ram:
## Purpose / Description

Add `observal support` CLI command group with `bundle` and `inspect` subcommands for generating and reviewing portable diagnostic archives that operators can attach to support tickets or GitHub issues.

**`observal support bundle`** produces a small, safe `.tar.gz` containing version info, sanitized config, health probes, aggregate table counts, error fingerprints, redacted logs, and system metrics. **`observal support inspect`** lets operators review bundle contents before sharing.

## Approach

- **Server-side collection**: New `POST /api/v1/support/collect` endpoint runs 6 collectors (versions, health, config, aggregates, errors, logs) with 10s timeouts. Preserves existing architecture: CLI → HTTP → API Server → {PG, CH, Redis}.
- **Local collectors**: System info and config allowlist filtering run in a thread pool on the operator's machine.
- **Redaction Layer**: Single chokepoint module — JWTs, AWS keys, URL credentials (any scheme via generic regex), high-entropy tokens, and sensitive JSON keys replaced with `<REDACTED>`. Entropy check skips already-redacted tokens to preserve URL structure.
- **Manifest**: SHA-256 file inventory for tamper detection. `node_id` is a truncated hash (not raw hostname). `flags_used["output"]` stores filename only.
- **Inspect**: Read-only via `tarfile`, path traversal protection via `os.path.normpath`, schema version warnings.
- **Error safety**: All exception messages use `type(exc).__name__` only — no credential leaks from SQLAlchemy connection errors.

## How Has This Been Tested?

**Unit/integration tests (268 pass, 1 skip):**
```bash
cd observal-server && uv run --with pytest --with pytest-asyncio --with pyyaml \
  --with typer --with rich --with docker --with hypothesis \
  pytest ../tests/test_support_redaction.py ../tests/test_support_manifest.py \
  ../tests/test_support_collectors.py ../tests/test_support_bundle_wiring.py \
  ../tests/test_support_inspect.py ../tests/test_support_integration.py \
  ../tests/test_support_pbt.py ../tests/test_support_endpoint.py \
  ../tests/test_cmd_support.py -v
```

**Manual CLI verification (requires `make rebuild` + `observal auth login`):**
```bash
# Generate and inspect a full bundle
observal support bundle --output test-bundle.tar.gz
observal support inspect test-bundle.tar.gz
observal support inspect test-bundle.tar.gz --show config/config.json
observal support inspect test-bundle.tar.gz --show aggregates/pg_table_counts.json

# Verify --no-include-system excludes system directory
observal support bundle --output no-system.tar.gz --no-include-system
observal support inspect no-system.tar.gz

# Error handling
observal support inspect nonexistent.tar.gz
observal support inspect test-bundle.tar.gz --show nonexistent.json

# Cleanup
rm test-bundle.tar.gz no-system.tar.gz
```

- Property-based testing (Hypothesis) is effective for validating redaction correctness — it found the entropy/URL interaction edge case where a redacted URL token could trigger the entropy rule.
- `tarfile.open(mode="r:gz")` with member filtering is the safe pattern for reading archives without extraction.
- `os.path.normpath` is the correct way to check path traversal — substring checks on `..` reject legitimate filenames.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,25 @@
 
 All notable changes to this project will be documented in this file.
 
+## [Unreleased]
+
+### Added — Support Bundle (`observal support`)
+
+New `observal support` CLI command group for generating and reviewing portable diagnostic archives. Operators can attach these to support tickets or GitHub issues to help debug deployment problems without exposing secrets or customer data.
+
+- **`observal support bundle`** — collects version info, sanitized config, health probes, aggregate table counts, recent error fingerprints, redacted structured logs, and optional system metrics into a `.tar.gz` archive
+  - All values pass through a single **Redaction Layer** before being written: JWTs, AWS keys, URL credentials, high-entropy tokens, and sensitive JSON keys are replaced with `<REDACTED>`
+  - Config filtering uses an explicit **allowlist** — keys not on the list are omitted entirely, not just redacted
+  - Server-side collectors run via `POST /api/v1/support/collect`; local collectors (system info, config filtering) run in parallel on the operator's machine
+  - Individual collector failures produce partial bundles (exit 0); total failure exits with code 1
+  - Archive permissions set to `0o600` (owner read/write only)
+  - Bundle manifest (`bundle_manifest.json`) includes schema version, collector results, redaction counts, and SHA-256 file inventory for tamper detection
+- **`observal support inspect <bundle.tar.gz>`** — opens a bundle read-only (never extracts to disk) and displays the manifest, a Rich file tree with human-readable sizes, and optionally prints a single file via `--show`
+  - Schema version warnings for bundles from newer CLI versions
+  - Path traversal protection on tar member filtering
+- No customer row data is ever included — only aggregate counts
+- GitHub issue template updated to ask for a support bundle attachment
+
 ## [0.4.0] - 2026-04-28
 
 ### Added
--- a/observal-server/api/routes/support.py
+++ b/observal-server/api/routes/support.py
@@ -1,0 +1,489 @@
+"""Server-side diagnostic collection endpoint for support bundles.
+
+Runs collectors for versions, health, config, aggregates, errors, and
+logs data. Each collector is wrapped in ``_run_collector`` with a
+10-second ``asyncio.wait_for`` timeout. Partial failures are reported
+in the response — the endpoint always returns 200 if at least one
+collector succeeds.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import hashlib
+import os
+import re
+import time
+from datetime import UTC, datetime, timedelta
+from typing import TYPE_CHECKING, Any
+
+import structlog
+from fastapi import APIRouter, Depends, Request
+from pydantic import BaseModel
+from sqlalchemy import text
+
+from api.deps import get_current_user, get_db
+from api.ratelimit import limiter
+from config import Settings, settings
+from services.clickhouse import CLICKHOUSE_DB, _query
+from services.redis import get_redis
+
+if TYPE_CHECKING:
+    from sqlalchemy.ext.asyncio import AsyncSession
+
+    from models.user import User
+
+logger = structlog.get_logger(__name__)
+
+router = APIRouter(prefix="/api/v1/support", tags=["support"])
+
+COLLECTOR_TIMEOUT_SECONDS = 10
+
+# Valid ClickHouse/PG table name: alphanumeric + underscores, starting with a letter or underscore
+_SAFE_TABLE_NAME_RE = re.compile(r"^[a-zA-Z_][a-zA-Z0-9_]*$")
+
+
+# ── Request / Response models ────────────────────────────────────────
+
+
+class CollectRequest(BaseModel):
+    collectors: list[str] = ["all"]
+    logs_since: str = "1h"
+
+
+class CollectorData(BaseModel):
+    ok: bool
+    duration_ms: int
+    data: Any = None
+    error: str | None = None
+
+
+class CollectResponse(BaseModel):
+    server_version: str
+    collectors: dict[str, CollectorData]
+
+
+# ── Collector wrapper ────────────────────────────────────────────────
+
+
+async def _run_collector(name: str, coro) -> tuple[str, CollectorData]:
+    """Run a single collector coroutine with a 10-second timeout.
+
+    Returns a (name, CollectorData) tuple regardless of success or failure.
+    """
+    start = time.monotonic()
+    try:
+        data = await asyncio.wait_for(coro, timeout=COLLECTOR_TIMEOUT_SECONDS)
+        elapsed_ms = int((time.monotonic() - start) * 1000)
+        return name, CollectorData(ok=True, duration_ms=elapsed_ms, data=data)
+    except TimeoutError:
+        return name, CollectorData(
+            ok=False,
+            duration_ms=COLLECTOR_TIMEOUT_SECONDS * 1000,
+            error=f"Collector timed out after {COLLECTOR_TIMEOUT_SECONDS}s",
+        )
+    except Exception as exc:
+        elapsed_ms = int((time.monotonic() - start) * 1000)
+        return name, CollectorData(ok=False, duration_ms=elapsed_ms, error=type(exc).__name__)
+
+
+# ── Individual collectors ────────────────────────────────────────────
+
+
+async def _collect_versions(db: AsyncSession) -> dict:
+    """Collect app version, build hash, Alembic revision, ClickHouse version + tables."""
+    result: dict[str, Any] = {}
+
+    # App version — try importlib.metadata first, fall back to hardcoded
+    try:
+        from importlib.metadata import version
+
+        result["app_version"] = version("observal-server")
+    except Exception:
+        result["app_version"] = "0.1.0"
+
+    # Build hash — read from BUILD_HASH env var or fall back to unknown
+    result["build_hash"] = os.environ.get("BUILD_HASH", "unknown")
+
+    # Alembic revision from PG
+    try:
+        row = await db.execute(text("SELECT version_num FROM alembic_version LIMIT 1"))
+        rev = row.scalar_one_or_none()
+        result["alembic_revision"] = rev or "unknown"
+    except Exception as exc:
+        result["alembic_revision"] = f"error: {type(exc).__name__}"
+
+    # ClickHouse version
+    try:
+        resp = await _query("SELECT version()")
+        if resp.status_code == 200:
+            result["clickhouse_version"] = resp.text.strip()
+        else:
+            result["clickhouse_version"] = f"error: HTTP {resp.status_code}"
+    except Exception as exc:
+        result["clickhouse_version"] = f"error: {type(exc).__name__}"
+
+    # ClickHouse table list
+    try:
+        resp = await _query(
+            "SELECT name FROM system.tables WHERE database = {db:String} FORMAT JSON",
+            {"param_db": CLICKHOUSE_DB},
+        )
+        if resp.status_code == 200:
+            rows = resp.json().get("data", [])
+            result["clickhouse_tables"] = [r["name"] for r in rows]
+        else:
+            result["clickhouse_tables"] = []
+    except Exception as exc:
+        result["clickhouse_tables"] = f"error: {type(exc).__name__}"
+
+    return result
+
+
+async def _collect_health(db: AsyncSession) -> dict:
+    """Run health probes against PG, CH, Redis, and OTEL collector."""
+    result: dict[str, Any] = {}
+
+    # PostgreSQL health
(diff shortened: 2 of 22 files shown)


observal_cli/support/redaction.py
```diff
@@ -0,0 +1,118 @@
...
+JWT_PATTERN = re.compile(r"eyJ[A-Za-z0-9_-]+\.[A-Za-z0-9_-]+\.[A-Za-z0-9_-]+")
+
+# AWS access key pattern
+AWS_KEY_PATTERN = re.compile(r"AKIA[0-9A-Z]{16}")
+
+# URL userinfo: scheme://user:password@host
+URL_USERINFO_PATTERN = re.compile(
+    r"((?:https?|postgresql\+asyncpg|postgres|redis|clickhouse)://)"
+    r"([^@/\s]+)@"
+)
```
Apoorvgarg-creator, line 30: **risky** the alternation only covers `postgresql+asyncpg`, `postgres`, `redis`, `clickhouse`, `https?`. Plain `postgresql://user:pass@host/db` (the most common form when DATABASE_URL comes from an env var or k8s secret) is not matched, so the credentials slip through. Verified locally:

```
>>> redact_string('postgresql://postgres:hunter2@db:5432/app')
('postgresql://postgres:hunter2@db:5432/app', 0)
```

Same for `mysql://`, `mongodb://`, `amqp://`, `kafka://`. Either widen the alternation, or use a generic `[a-z][a-z0-9+\-.]*://` and rely on the userinfo shape to gate it. Given DATABASE_URL is in the allowlist, this is the path most likely to leak.

observal_cli/cmd_support.py
```diff
@@ -0,0 +1,573 @@
...
+
+
+# ── CLI version helper ───────────────────────────────────────────────
+
+
+def _get_cli_version() -> str:
+    try:
+        from importlib.metadata import version as pkg_version
+
+        return pkg_version("observal")
```
Apoorvgarg-creator: **bug** the distribution name in pyproject.toml is `observal-cli`, not `observal`, so this lookup always raises and we fall through to `"dev"`. The bundle's `cli_version` will be `"dev"` for every real install. Same problem on the server side at `_collect_versions` (`version("observal")` vs `observal-server`). Use the actual dist name.

observal-server/api/routes/support.py
```diff
@@ -0,0 +1,508 @@
...
+
+async def _collect_versions(db: AsyncSession) -> dict:
+    """Collect app version, build hash, Alembic revision, ClickHouse version + tables."""
+    result: dict[str, Any] = {}
+
+    # App version — try importlib.metadata first, fall back to hardcoded
+    try:
+        from importlib.metadata import version
+
+        result["app_version"] = version("observal")
```
Apoorvgarg-creator: **bug** `version("observal")` will not resolve since the installed dist is `observal-server`. Always falls back to `"0.1.0"`, which silently misreports the running version on every bundle. Same hardcoded fallback again at line 506. Pull from the right dist or read from a `__version__` constant.

observal-server/api/routes/support.py
```diff
@@ -0,0 +1,508 @@
...
+        result["redis"] = {"status": "ok" if pong else "error", "latency_ms": redis_ms}
+    except Exception as exc:
+        redis_ms = int((time.monotonic() - redis_start) * 1000)
+        result["redis"] = {"status": "error", "latency_ms": redis_ms, "error": str(exc)}
+
+    # OTEL collector health (best-effort HTTP GET)
+    otel_start = time.monotonic()
+    try:
+        async with httpx.AsyncClient(timeout=5.0) as client:
+            otel_resp = await client.get("http://otel-collector:13133/")
```
Apoorvgarg-creator: **bug** the OTEL collector was removed in 0.4.0 (CHANGELOG: `remove OTEL collector, add nginx /v1/ routing`). This will always ConnectError and write `unavailable` into every bundle. Either drop the probe entirely or repoint at the new ingest path.

observal_cli/support/manifest.py
```diff
@@ -0,0 +1,75 @@
...
+    sha256: str
+
+
+@dataclass
+class BundleManifest:
+    bundle_schema_version: str = "1"
+    created_at: str = ""
+    cli_version: str = ""
+    host_os: str = ""
+    node_id: str = ""  # socket.gethostname() — identifies which machine produced this bundle
```
Apoorvgarg-creator, line 26: **risky** Requirement 2.9 in the proposal says no hostnames anywhere, but `node_id = socket.gethostname()` is written into the manifest unconditionally. Hostnames in production often encode env/region/team (`prod-us-east-api-7`) which is exactly what the spec says not to leak. Drop it, or hash it (sha256 truncated) if you actually want stable per-host correlation.

observal_cli/cmd_support.py
```diff
@@ -0,0 +1,573 @@
...
+            collector_summary[r.name]["error"] = r.error
+
+    manifest = BundleManifest(
+        bundle_schema_version="1",
+        created_at=datetime.now(UTC).isoformat(),
+        cli_version=cli_version,
+        host_os=platform.system(),
+        node_id=socket.gethostname(),
+        flags_used={
+            "output": str(output),
```
Apoorvgarg-creator: **risky** `flags_used["output"] = str(output)` writes the full filesystem path into the manifest, which on most desktops contains the operator's username (`/Users/alice/...`, `/home/alice/...`). Either store only `output.name` or drop it from the manifest, since the filename is already the bundle's own name.

CHANGELOG.md
```diff
@@ -1,888 +1,907 @@
...
-
-### Ui
-
-- brighter borders, responsive text scaling, observal favicon ([909a9b6](https://github.com/BlazeUp-AI/Observal/commit/909a9b6dd7cedc9ddaff31b208f09ace9d35be7e))
-
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+## [Unreleased]
```
Apoorvgarg-creator: **bug** the whole file got rewritten with CRLF line endings and the unicode arrows in older entries are now mojibake (`ΓåÆ` instead of `→`). That's why this diff is +907/-888 instead of a small `[Unreleased]` block. Re-save with LF line endings (`dos2unix CHANGELOG.md`) and restore the arrows so the diff is the few lines you actually intended to add.

observal-server/api/routes/support.py
```diff
@@ -0,0 +1,508 @@
...
+    # Build hash — read from BUILD_HASH env var or fall back to unknown
+    result["build_hash"] = os.environ.get("BUILD_HASH", "unknown")
+
+    # Alembic revision from PG
+    try:
+        row = await db.execute(text("SELECT version_num FROM alembic_version LIMIT 1"))
+        rev = row.scalar_one_or_none()
+        result["alembic_revision"] = rev or "unknown"
+    except Exception as exc:
+        result["alembic_revision"] = f"error: {exc}"
```
Apoorvgarg-creator: **risky** writing `f"error: {exc}"` straight into the bundle is a potential leak vector: SQLAlchemy connection errors typically include the connection URL with credentials. Wrap as `error: {type(exc).__name__}` or run the string through the redaction layer before storing. Same shape at lines 124, 138, 257, 282, 287.

observal_cli/cmd_support.py
```diff
@@ -0,0 +1,573 @@
...
+    for member in sorted(members, key=lambda m: m.name):
+        if member.isfile():
+            size = _human_size(member.size)
+            tree.add(f"{member.name}  [dim]{size}[/dim]")
+    console.print(tree)
+
+
+def _is_safe_tar_member(member: tarfile.TarInfo) -> bool:
+    """Reject tar members with path traversal attacks."""
+    return not member.name.startswith(("/", "..")) and ".." not in member.name
```
Apoorvgarg-creator: **nit** `".." not in member.name` rejects any path containing a `..` substring, including legitimate names like `foo..bar.json`. The startswith check plus a normalized-path check (`os.path.normpath(name).startswith(("..", "/"))`) is the usual pattern.

observal_cli/cmd_support.py
```diff
@@ -0,0 +1,573 @@
...
+
+            local_tasks.append(_run_system_collector)
+        except ImportError:
+            pass
+
+    with ThreadPoolExecutor(max_workers=4) as pool:
+        futures = [pool.submit(fn) for fn in local_tasks]
+        for future in futures:
+            try:
+                result = future.result(timeout=10)
```
Apoorvgarg-creator, line 321: **nit** `future.result(timeout=10)` only stops waiting, it does not kill the worker thread. If a collector hangs (e.g. a blocking syscall in `_get_memory_total`) the thread leaks for the lifetime of the process. For these short collectors it's fine in practice, just worth a comment so future readers don't assume the timeout cancels work.

observal_cli/cmd_support.py
```diff
@@ -0,0 +1,573 @@
...
+            # Log lines: redact each line individually, write as newline-delimited JSON
+            if isinstance(result.data, dict):
+                lines = result.data.get("lines", [])
+                redacted_lines: list[str] = []
+                for line in lines:
+                    redacted_line, line_count = redact_value(line)
+                    redaction_stats.record("logs/recent.log", line_count)
+                    redacted_lines.append(json.dumps(redacted_line, default=str))
+                if redacted_lines:
+                    files["logs/recent.log"] = "\n".join(redacted_lines).encode("utf-8")
```
Apoorvgarg-creator: **nit** the file ends up as newline-delimited JSON but the name is `recent.log`. Either rename to `recent.ndjson` or drop the per-line `json.dumps` and write the original strings, otherwise tools that read `*.log` will trip on the JSON wrapping.

observal-server/api/routes/support.py
```diff
@@ -0,0 +1,508 @@
...
+    "errors": lambda db, logs_since: _collect_errors(db),
+    "logs": lambda db, logs_since: _collect_logs(logs_since),
+}
+
+
+# ── Endpoint ─────────────────────────────────────────────────────────
+
+
+@router.post("/collect", response_model=CollectResponse)
+@limiter.limit("5/minute")
```
Apoorvgarg-creator, line 457: **doubt** `get_current_user` means any authenticated user can dump aggregate row counts, full server config (allowlisted), and error fingerprints across the whole deployment. The proposal does say `SHALL NOT require admin role`, so this is intentional, but on a multi-tenant install this is a footgun. Worth confirming this is desired vs. gating behind an admin check.

observal_cli/support/collectors.py
```diff
@@ -0,0 +1,107 @@
...
+    """Detect if running inside a container.
+
+    Checks for Docker (/.dockerenv) and Podman (/run/.containerenv).
+    Returns the runtime name or None if not in a container.
+    """
+    if os.path.exists("/.dockerenv"):
+        return "docker"
+    if os.path.exists("/run/.containerenv"):
+        return "podman"
+    return None
```
Apoorvgarg-creator, line 110: **nit** Kubernetes pods do not have `/.dockerenv`. Adding `os.environ.get("KUBERNETES_SERVICE_HOST")` as a third branch (returning `"kubernetes"`) would catch the most common self-hosted deployment shape.

Apoorvgarg-creator (changes requested):
Tests are green and the inspect flow works end-to-end, but a few of these need to land before merge (postgresql:// creds leak, hostname in manifest, dead otel-collector probe, mangled CHANGELOG). Rest are nits/doubts. Once the bugs are fixed, LGTM.

naraen-ram:
### Review feedback addressed

- Widened URL userinfo regex to `[a-z][a-z0-9+\-.]*://` (catches `postgresql://`, `mysql://`, `mongodb://`, etc.)
- Fixed package names in `importlib.metadata.version()` (`observal-cli` / `observal-server`)
- Removed dead OTEL collector probe (removed in v0.4.0)
- Fixed CHANGELOG CRLF line endings and unicode mojibake
- Hashed `node_id` (SHA-256 truncated to 12 chars) instead of raw hostname
- Renamed `logs/recent.log` → `logs/recent.ndjson` to match NDJSON content
- Added Kubernetes detection (`KUBERNETES_SERVICE_HOST` env var)

observal-server/api/routes/support.py
naraen-ram: This is intentional for v1. The spec says "SHALL NOT require admin role" (Requirement 1.6) so that any operator experiencing an issue can generate a bundle without needing elevated access. The data exposed is limited to aggregate counts, allowlisted config keys (with credentials redacted), and error fingerprints (no exception messages).

That said, your multi-tenant concern is valid. Today Observal is single-tenant (one org per deployment), so this is safe. If multi-tenant support lands, this endpoint should be scoped to the user's organization.

Haz3-jolt (approved):

Issue #617, opened 2026-04-28
DB: Agent version tables + schema restructure


Haz3-jolt:
**Parent:** #615 (Registry Update v1.0)
**Depends on:** #616
**Phase:** 1

## Summary

Restructure `agents` table into identity-only + `agent_versions` table. Agent becomes a package identity record; version-specific data lives in `agent_versions`.

## `agents` table changes

**Keeps:** `id`, `name`, `owner`, `created_by`, `owner_org_id`, `visibility` (enum: public/private), `created_at`, `updated_at`

**Adds:**
- `latest_version_id` — FK to `agent_versions.id`
- `co_maintainers` — JSON list of user UUIDs (Phase 1 simplicity; migrates to join table in Phase 2)

**Removes (moved to `agent_versions`):** `version`, `description`, `prompt`, `model_name`, `model_config_json`, `external_mcps`, `supported_ides`, `required_ide_features`, `inferred_supported_ides`, `status`, `rejection_reason`, `download_count`, `unique_users`

**Note:** No `git_url` on agents table. Registry DB is the source of truth, not git repos.

## `agent_versions` table (new)

| Column | Type | Notes |
|---|---|---|
| `id` | UUID | PK |
| `agent_id` | UUID | FK to `agents.id` |
| `version` | String(50) | Semver string |
| `description` | Text | |
| `prompt` | Text | System prompt |
| `model_name` | String(100) | |
| `model_config_json` | JSON | |
| `external_mcps` | JSON | |
| `supported_ides` | JSON | |
| `required_ide_features` | JSON | |
| `inferred_supported_ides` | JSON | |
| `yaml_snapshot` | Text | Complete `observal-agent.yaml` at this version |
| `ide_configs` | JSON | Generated IDE config files per IDE, keyed by IDE identifier. Stored at publish time, served at pull time. |
| `lock_snapshot` | Text | Resolved lock file content at this version (with integrity hashes) |
| `status` | Enum | `draft`, `pending`, `active`, `rejected` |
| `is_prerelease` | Boolean | True for beta releases (Phase 3) |
| `promoted_from` | UUID | Nullable FK to `agent_versions.id` (Phase 3) |
| `rejection_reason` | Text | Nullable |
| `download_count` | Integer | Per-version |
| `released_by` | UUID | FK to `users.id` |
| `released_at` | DateTime | |
| `reviewed_by` | UUID | Nullable |
| `reviewed_at` | DateTime | Nullable |
[...]

PR #653, opened 2026-04-30
feat(db): listing tables identity-only cleanup

### LINKED ISSUE #616: DB: Component version tables for all 5 types
**Parent:** #615 (Registry Update v1.0)
**Phase:** 1

## Summary

Create `{type}_versions` tables for all 5 component types (MCP, skill, hook, prompt, sandbox). Transform existing single-row-per-item into identity + version tables.

### Pattern: `{type}_listings` (identity) + `{type}_versions` (new)

Applied to: `mcp_listings`, `skill_listings`, `hook_listings`, `prompt_listings`, `sandbox_listings`

**Identity table keeps:** `id`, `name`, `owner`, `category`, `submitted_by`, `owner_org_id`, `is_private`, `created_at`, `updated_at`

**Identity table adds:**
- `latest_version_id` — FK to `{type}_versions.id`

**Identity table removes (moved to versions):** `version`, `description`, type-specific content fields, `status`, `rejection_reason`, `download_count`, `changelog`

### `{type}_versions` table schema

| Column | Type | Notes |
|---|---|---|
| `id` | UUID | PK |
| `listing_id` | UUID | FK to `{type}_listings.id` |
| `version` | String(50) | Semver string |
| `description` | Text | Version-specific description |
| `content` | Text/JSON | Inline content (skills, hooks, prompts) OR null for large types |
| `source_url` | String(500) | Optional — external source URL (git repo, package registry). Used for MCP servers and sandboxes whose code lives externally. |
| `source_ref` | String(255) | Optional — git tag/SHA/branch at the external source |
| `resolved_sha` | String(40) | Resolved commit SHA at publish time. Prevents tag-mutability attacks. Required when `source_url` is set. |
| `status` | Enum | `draft`, `pending`, `approved`, `rejected` |
| `rejection_reason` | Text | Nullable |
| `download_count` | Integer | Per-version |
| `changelog` | Text | What changed |
| `released_by` | UUID | FK to `users.id` |
| `released_at` | DateTime | |
| `reviewed_by` | UUID | Nullable |
| `reviewed_at` | DateTime | Nullable |
| `created_at` | DateTime | |

**Unique constraint:** `(listing_id, version)`

**New column vs old spec:** `resolved_sha` — per design review, we resolve `source_ref` (which could be a mutable tag) to a full commit SHA at publish time and store it. This prevents an approved version from silently pointing at different code if the upstream tag gets force-pushed.

**Type-specific extra columns:**

| Type | Extra columns |
|---|---|
| MCP | `transport`, `framework`, `docker_image`, `command`, `args`, `url`, `headers`, `auto_approve`, `mcp_validated`, `tools_schema`, `environment_variables`, `supported_ides`, `setup_instructions` |
| Skill | `supported_ides`, `skill_type`, `trigger_pattern` |
| Hook | `supported_ides`, `hook_type`, `trigger_event` |
| Prompt | `supported_ides` |
| Sandbox | `docker_image`, `supported_ides` |

## Content Storage Model

| Type | Typical size | Storage | What version row holds |
|---|---|---|---|
| MCP servers | Full application | External git repo | Metadata + `source_url`/`source_ref`/`resolved_sha` |
| Sandboxes | Docker configs | External git repo | Metadata + `source_url`/`source_ref`/`resolved_sha` |
| Skills | Markdown / small script | DB inline | Full content in `content` column |
| Hooks | Few lines of config | DB inline | Full content in `content` column |
| Prompts | Text string | DB inline | Full content in `content` column |

**Key distinction:** MCP servers and sandboxes are real software — they live in their own repos and are installed via `npx`, `pip`, `docker pull`, etc. Observal doesn't store or serve their source code. It version-tracks the *reference* (which repo, which commit SHA) so the lock file can pin a specific version and consumers get reproducible installs.

## External Source Integrity (per design review)

When a version is published with `source_url` + `source_ref`:
1. Server resolves `source_ref` to a commit SHA via git hosting API (GitHub/GitLab)
2. Stores the SHA in `resolved_sha`
3. Lock file entries for this component include `source_sha: <resolved_sha>`
4. On `agent pull`, CLI can verify the SHA hasn't changed (warn, don't block)

This is cheap to implement and prevents the entire class of tag-mutability attacks.

## Migration Steps

1. Create all 5 `{type}_versions` tables (with `resolved_sha` column)
2. For each existing listing row: create one version row with current data
3. Add `latest_version_id` FK to each listing table, set to newly created version
4. Drop version-specific columns from listing tables

## Files to modify

- `models/mcp.py`, `models/skill.py`, `models/hook.py`, `models/prompt.py`, `models/sandbox.py`
- New migration: `alembic/versions/00XX_component_version_tables.py`
- `schemas/` for all component types (request/response shapes)

Haz3-jolt:
## Purpose / Description

Completes the listing identity-only cleanup started in #650. Transforms all 5 listing models to identity-only by dropping version-specific columns, adding backward-compatible `@property` delegates, and updating all SQL queries and submission routes to work through version tables.

## Fixes
* Closes #616

## Approach

- Dropped version-specific columns from all 5 listing tables via migration 0022
- Added backward-compatible `@property` delegates on listing models that delegate to `latest_version` (same pattern as Agent from #617)
- Compat property setters raise `RuntimeError` if `latest_version` is None (prevents silent data loss on state transitions)
- Rewrote submission routes to create Listing + Version separately with `db.flush()` between
- Updated all SQL queries that referenced removed columns (status, description) to join through version tables
- Removed `source_url`/`source_ref`/`resolved_sha` from skill, hook, and prompt version tables — only MCP and sandbox have external git repos per the epic design
- Updated tests to check columns on Version models and handle the two-object creation pattern

## How Has This Been Tested?

- 1859 root-level tests pass
- 133 server-internal tests pass
- Ruff lint/format clean
- Schema redesign tests updated (columns now on version models, inline types verified to not have source fields)
- Registry type tests updated (listing+version creation pattern)
- Environment detection tests updated (mock `latest_version` relationship for refresh)

## Checklist
- [x] All commits are signed off (`git commit -s`) per the DCO
- [x] You have a descriptive commit message with a short title
- [ ] UI changes: include screenshots

### DIFF
--- a/observal-server/alembic/versions/0022_listing_identity_cleanup.py
+++ b/observal-server/alembic/versions/0022_listing_identity_cleanup.py
@@ -1,0 +1,146 @@
+"""Drop version-specific columns from listing tables (identity-only cleanup).
+
+Revision ID: 0022
+Revises: 0021
+Create Date: 2026-04-30
+"""
+
+from collections.abc import Sequence
+from typing import Union
+
+from alembic import op
+
+revision: str = "0022"
+down_revision: Union[str, Sequence[str], None] = "0021"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    # --- mcp_listings ---
+    op.drop_index("ix_mcp_listings_status", table_name="mcp_listings")
+    for col in [
+        "version",
+        "git_url",
+        "git_ref",
+        "description",
+        "transport",
+        "framework",
+        "docker_image",
+        "command",
+        "args",
+        "url",
+        "headers",
+        "auto_approve",
+        "mcp_validated",
+        "tools_schema",
+        "environment_variables",
+        "supported_ides",
+        "setup_instructions",
+        "changelog",
+        "status",
+        "rejection_reason",
+        "download_count",
+    ]:
+        op.drop_column("mcp_listings", col)
+
+    # --- skill_listings ---
+    op.drop_index("ix_skill_listings_status", table_name="skill_listings")
+    for col in [
+        "version",
+        "description",
+        "git_url",
+        "git_ref",
+        "supported_ides",
+        "status",
+        "rejection_reason",
+        "download_count",
+        "skill_path",
+        "target_agents",
+        "task_type",
+        "triggers",
+        "slash_command",
+        "has_scripts",
+        "has_templates",
+        "is_power",
+        "power_md",
+        "mcp_server_config",
+        "activation_keywords",
+    ]:
+        op.drop_column("skill_listings", col)
+
+    # --- hook_listings ---
+    op.drop_index("ix_hook_listings_status", table_name="hook_listings")
+    for col in [
+        "version",
+        "description",
+        "git_url",
+        "git_ref",
+        "supported_ides",
+        "status",
+        "rejection_reason",
+        "download_count",
+        "event",
+        "execution_mode",
+        "priority",
+        "handler_type",
+        "handler_config",
+        "input_schema",
+        "output_schema",
+        "scope",
+        "tool_filter",
+        "file_pattern",
+    ]:
+        op.drop_column("hook_listings", col)
+
+    # --- prompt_listings ---
+    op.drop_index("ix_prompt_listings_status", table_name="prompt_listings")
+    for col in [
+        "version",
+        "description",
+        "git_url",
+        "git_ref",
+        "supported_ides",
+        "status",
+        "rejection_reason",
+        "download_count",
+        "category",
+        "template",
+        "variables",
+        "model_hints",
+        "tags",
+    ]:
+        op.drop_column("prompt_listings", col)
+
+    # --- sandbox_listings ---
+    op.drop_index("ix_sandbox_listings_status", table_name="sandbox_listings")
+    for col in [
+        "version",
+        "description",
+        "git_url",
+        "git_ref",
+        "supported_ides",
+        "status",
+        "rejection_reason",
+        "download_count",
+        "runtime_type",
+        "image",
+        "dockerfile_url",
+        "resource_limits",
+        "network_policy",
+        "allowed_mounts",
+        "env_vars",
+        "entrypoint",
+    ]:
+        op.drop_column("sandbox_listings", col)
+
+    # --- Drop source fields from inline-only version tables ---
+    # Skills, hooks, and prompts are inline content (no git repos).
+    # source_url/source_ref/resolved_sha only belong on MCP and sandbox.
+    for table in ["skill_versions", "hook_versions", "prompt_versions"]:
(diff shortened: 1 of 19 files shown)


Haz3-jolt:
Adversarial review.

The silent no-op pattern on all compat property setters is the real concern here. Every setter does "if self.latest_version: self.latest_version.status = value" which means if latest_version is None for any reason (stale session, race, manual DB insert, relationship not loaded), the write is silently discarded. The review workflow sets listing.status = approved, gets a 200 back, and nothing actually changed in the DB. This is nearly impossible to debug from the user side. Either raise RuntimeError in the setter when latest_version is missing, or at minimum log a warning. Silent data loss on a state machine transition is a bad failure mode.

The redundant eager loading is worth noting. Each listing now fires two selectin queries: one for versions (full history) and one for latest_version (single row that already exists in the versions collection). At list-endpoint scale with 20-50 listings this doubles the secondary query count for no benefit. Consider deriving latest_version from the versions list via a python property, or make versions lazy since only specific endpoints need full history.

Three flushes per submission (listing flush, version flush, FK update, commit) is fine for correctness but worth knowing about. You could pre-generate the version UUID in Python (you already have default=uuid.uuid4) and set listing.latest_version_id in the same flush as the version insert. Minor optimization, not blocking.

Category placement inconsistency: McpListing.category is a real column (identity-level data), but PromptVersion.category is version-level with a compat delegate back to the listing. This is probably intentional per the design (MCP category is fixed across versions, prompt category can change per version), but a code comment explaining why would save the next person 10 minutes of confusion.

No changelog setter on McpListing. The property is read-only. If any existing code path assigns listing.changelog it will raise AttributeError rather than silently no-op. This is actually better behavior than the silent swallow, but check callers.

None of these block merge except the silent setter issue. That one will bite you the first time a review action hits a listing in a weird state.

Haz3-jolt:
Adversarial review round 2. The RuntimeError fix landed correctly on McpListing, PromptListing, and SandboxListing but HookListing was missed entirely. Every setter in HookListing still uses the silent if self.latest_version pattern instead of raising when latest_version is None. That means the review workflow approving or rejecting a hook will silently no-op under the same edge conditions you fixed for the other four types.

SkillListing is mostly correct but three setters at the bottom still use the old silent pattern: power_md, mcp_server_config, and activation_keywords. The rest of SkillListing raises properly so this looks like copy-paste that stopped short.

The source_url/source_ref/resolved_sha removal from skill, hook, and prompt versions is a good design call. Only MCP and sandbox reference external git repos so those fields had no business on inline content types. The migration dropping them and the test parametrization splitting external vs inline are both correct.

The getattr fallback in agent_resolver.py for git_url is fine as a transitional safety net since not all component types will have that attribute anymore. Once the version-aware API layer lands that can become an explicit check on the version model rather than getattr.

Fix the HookListing setters and the three SkillListing stragglers and this is ready to merge.

Haz3-jolt:
Finished multiple adversarial reviews

Issue #622, opened 2026-04-28
API: Agent versioning endpoints (release, review, diff)


Haz3-jolt:
## Parent Epic
Part of #615 — Registry Update v1.0
**Phase:** 1

## Overview

Implement the agent-specific versioning API endpoints: creating releases, reviewing versions, serving IDE configs, diffing between versions, and fetching version details.

## Dependencies
- #617 (DB: Agent version tables)
- #620 (Service: Agent config generation)

### New Endpoints

| Method | Path | Description | Auth |
|---|---|---|---|
| GET | `/api/v1/agents/{id}/versions` | List all versions | Any authed |
| GET | `/api/v1/agents/{id}/versions/{version}` | Get specific version | Any authed |
| POST | `/api/v1/agents/{id}/versions` | Create new release | Owner/co-maintainer |
| POST | `/api/v1/agents/{id}/versions/{version}/review` | Approve/reject | Reviewer+ |
| GET | `/api/v1/agents/{id}/versions/{version}/ide/{ide}` | Get generated IDE config for a specific version and IDE | Any authed |
| GET | `/api/v1/agents/{id}/versions/{v1}/diff/{v2}` | YAML diff between versions | Owner/co-maintainer/reviewer |

### Modified Endpoints

| Method | Path | Changes |
|---|---|---|
| GET | `/api/v1/agents/{id}` | Returns agent identity + latest approved version |
| POST | `/api/v1/agents` | Creates agent identity + initial version |
| PUT | `/api/v1/agents/{id}` | Updates identity fields only (name, visibility) |

#### `POST /api/v1/agents/{id}/versions` — Create Release

```json
{
  "version": "1.3.0",
  "description": "Added security scanning capabilities",
  "prompt": "You are a code review agent focused on security...",
  "model_name": "claude-sonnet-4",
  "model_config_json": {"max_tokens": 4096, "temperature": 0.2},
  "external_mcps": [...],
  "supported_ides": ["claude-code", "cursor", "kiro"],
  "yaml_snapshot": "name: my-agent\nversion: 1.3.0\n...",
  "is_prerelease": false,
  "components": [
    {"type": "mcp", "name": "filesystem-server", "resolved_version": "1.2.3", "component_id": "uuid"},
    {"type": "skill", "name": "code-review", "resolved_version": "2.0.7", "component_id": "uuid"}
  ]
}
```

On receiving this request, the server:
1. Validates semver format
[...]

PR #655, opened 2026-04-30
feat(api): agent versioning endpoints with IDE config and YAML diff

Haz3-jolt:
## Purpose/Description

Add agent-specific versioning API endpoints that go beyond the generic component version factory. Agents have richer version data (components, goal templates, IDE configs, YAML snapshots) requiring specialized endpoints.

## Fixes

Refs: #622

## Approach

- New `api/routes/agent_versions.py` with 6 endpoints mounted on the existing `/api/v1/agents` router
- Standalone async functions exposed for direct unit testing (same pattern as `component_versions.py`)
- Semver comparison on approval prevents `latest_version_id` regression
- IDE config endpoint returns cached configs or generates on-the-fly
- YAML diff uses `difflib.unified_diff` with fallback to structural JSON comparison when `yaml_snapshot` is None
- New schemas: `AgentVersionCreateRequest`, `AgentVersionReviewRequest` in `schemas/agent.py`

### Endpoints

| Method | Path | Description |
|--------|------|-------------|
| GET | `/agents/{id}/versions` | List all versions (paginated) |
| GET | `/agents/{id}/versions/{version}` | Get specific version detail |
| POST | `/agents/{id}/versions` | Create new agent version |
| POST | `/agents/{id}/versions/{version}/review` | Approve/reject (reviewer role) |
| GET | `/agents/{id}/versions/{version}/ide/{ide}` | Serve IDE config |
| GET | `/agents/{id}/versions/{v1}/diff/{v2}` | Unified YAML diff |

## How Has This Been Tested?

- 20 unit tests in `tests/test_agent_versions_api.py` covering:
  - List versions (empty, with data, pagination, agent not found)
  - Get version (found, not found)
  - Create version (happy path, bad semver, duplicate 409, not owner 403)
  - Review version (approve updates latest, reject stores reason, non-pending 422, not found 404)
  - IDE config (cached hit, on-the-fly generation)
  - YAML diff (snapshot diff, structural fallback)
- `ruff check` and `ruff format` pass

## Checklist

- [x] Code follows project conventions
- [x] Tests added
- [x] Lint passes
- [x] Single squashed commit with DCO sign-off

### DIFF
--- a/observal-server/api/routes/agent.py
+++ b/observal-server/api/routes/agent.py
@@ -1407,3 +1407,8 @@ async def submit_draft(
     return _agent_to_response(
         agent, name_map, created_by_email=current_user.email, created_by_username=current_user.username
     )
+
+
+from api.routes.agent_versions import agent_version_router  # noqa: E402
+
+router.include_router(agent_version_router)
--- a/observal-server/api/routes/agent_versions.py
+++ b/observal-server/api/routes/agent_versions.py
@@ -1,0 +1,542 @@
+"""Version-specific endpoints for agents.
+
+Mounted on the /api/v1/agents router via::
+
+    router.include_router(agent_version_router)
+
+All paths are relative to /api/v1/agents (no extra prefix).
+"""
+
+from __future__ import annotations
+
+import difflib
+import json
+from datetime import UTC, datetime
+
+from fastapi import APIRouter, Depends, HTTPException, Query
+from sqlalchemy import func, select
+from sqlalchemy.ext.asyncio import AsyncSession  # noqa: TC002
+
+from api.deps import get_db, get_effective_agent_permission, require_role
+from models.agent import (
+    Agent,
+    AgentGoalSection,
+    AgentGoalTemplate,
+    AgentStatus,
+    AgentVersion,
+)
+from models.mcp import McpListing
+from models.skill import SkillListing
+from models.user import User, UserRole
+from schemas.agent import (  # noqa: TC001
+    AgentVersionCreateRequest,
+    AgentVersionReviewRequest,
+)
+from services.agent_config_generator import generate_agent_config
+from services.agent_resolver import validate_component_ids
+from services.audit_helpers import audit
+from services.ide_feature_inference import compute_supported_ides, infer_required_features
+from services.versioning import parse_semver, validate_semver
+
+agent_version_router = APIRouter()
+
+# Import _load_agent from agent.py to avoid duplication.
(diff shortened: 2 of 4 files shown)


Haz3-jolt:
Adversarial review.

Route path collision between the diff endpoint and the get-version endpoint. GET /{agent_id}/versions/{version} and GET /{agent_id}/versions/{v1}/diff/{v2} both match the same path structure. When someone hits GET /agents/abc/versions/1.0.0/diff/2.0.0, FastAPI will try to match /{agent_id}/versions/{version} first with version=1.0.0 and the remaining /diff/2.0.0 will 404, or it will match with version=1.0.0/diff/2.0.0 depending on path resolution order. In practice FastAPI resolves by declaration order so since get_version is declared first (line 482) it will capture the request before the diff route (line 546) gets a chance. The diff endpoint is unreachable unless it is declared before the single-version GET, or the path is restructured to something unambiguous like /{agent_id}/diff/{v1}/{v2}.

Actually wait, FastAPI path parameters only match a single segment by default, so /versions/1.0.0/diff/2.0.0 has 4 segments while /{agent_id}/versions/{version} expects 3 segments. So FastAPI will correctly match the diff route. Disregard the collision concern, the routes are fine.

The real issue is the semver comparison in the approval logic calls parse_semver three times for the same two strings. Minor but the parse_semver(ver.version) is called once for the None check and would be called again for the comparison if Python short-circuits differently. Store the results in locals.

The _create_agent_version function does db.flush() after adding the version, then adds AgentComponent rows, then optionally adds GoalTemplate rows with another flush, then does a final commit. If any of the component or goal creation fails after the first flush, the version row exists in the transaction but the commit never happens so the transaction rolls back cleanly. That is correct.

The IDE config endpoint generates config on the fly if not cached but never persists the generated result back to ver.ide_configs. So every uncached request regenerates from scratch. If this is intentional (cache written at release time only, on-the-fly is a fallback for draft/preview), fine. If not, you are missing an await db.commit() after writing back to ver.ide_configs.

The diff endpoint returns raw diff_lines as a list of strings including newline characters. The consumer will get lines like ["--- v1.0.0\n", "+++ v2.0.0\n", ...]. This is usable but the trailing newlines in each element are slightly awkward for JSON consumers. Consider stripping them or joining into a single string. Non-blocking, just noting the API ergonomics.

The late import of _load_agent inside a wrapper function is fine for avoiding circular imports but adds one function call of indirection per request. Not a performance concern at your scale but worth knowing it is there.

Overall this is solid. No blocking issues.

PR #681, opened 2026-05-01
fix(migrate): handle cross-version import and include all ClickHouse tables in deep copy

Lokesh7025:
1. Filter archive columns to target schema -- extra columns from newer schemas are silently skipped instead of causing insert failures.

2. Schema version mismatch is now a warning, not a hard exit -- allows importing archives from newer Observal versions into older targets.

3. Validate stale .import_state.json -- if a "completed" table is actually empty (e.g., volumes were nuked after a prior import), the resume state is invalidated and the table is re-imported.

4. Backfill NULL owner_org_id after import -- agents and listings imported from older archives get their org inferred from the creator's user record, preventing them from being invisible in org-scoped API queries.

5. Add --project-id flag to import-telemetry that rewrites the project_id column in Parquet files before inserting, fixing invisible traces when source/target orgs differ.

6. Add otel_logs, security_events, and webhook_deliveries to CLICKHOUSE_TABLES so deep copy covers all ClickHouse tables. otel_logs powers the frontend Traces/Sessions page and was previously lost on volume nuke + rebuild.

## Purpose / Description
When migrating Observal across schema versions (e.g., main to v0.4.0), the shallow copy import would fail on column mismatches, hard-exit on version checks, skip tables due to stale resume state, and leave agents invisible due to NULL owner_org_id. The deep copy was also missing 3 of 9 ClickHouse tables, most critically otel_logs which powers the entire Traces/Sessions frontend page. This PR fixes all six issues so that export + import preserves all data without manual SQL intervention.

## Fixes
* Fixes #

## Approach
1. **Column filtering** (`_insert_table`): Intersect archive columns with target schema columns. Extra columns are logged and skipped.
2. **Schema version warning** (`_import_archive`): Changed version mismatch from `raise typer.Exit(1)` to a yellow warning appended to the result's warnings list.
3. **Resume state validation** (`_import_telemetry`): After loading `.import_state.json`, each "completed" table is checked for actual data. Empty tables get their resume state cleared.
4. **owner_org_id backfill** (`_import_archive`): After all table inserts, runs UPDATE on agents and listing tables to set `owner_org_id` from the creator's user record where NULL.
5. **--project-id flag** (`import-telemetry`): New `_rewrite_project_id` helper uses pyarrow to rewrite the `project_id` column in Parquet files before ClickHouse insert. Needed because `project_id` is a sort key and cannot be UPDATEd after insertion.
6. **Missing tables** (`CLICKHOUSE_TABLES`): Added `otel_logs` (time_col: `Timestamp`), `security_events`, and `webhook_deliveries` so deep copy covers all 9 ClickHouse tables.

## How Has This Been Tested?

- All 191 migration tests pass (`test_migrate.py` + `test_migrate_telemetry.py`)
- Full test suite: 2022 passed (20 pre-existing failures in `test_git_mirror`, `test_uninstall`, `test_cli_errors` unrelated to migration)
- ruff check and ruff format pass
- Manually tested end-to-end: exported from internal.observal.io running main (schema 0022), imported into a fresh v0.4.0 instance (schema 0018). All 13 agents visible in frontend after import.

- ClickHouse ReplacingMergeTree sort key columns (like `project_id`) cannot be UPDATEd after insert. To change them you need to CREATE AS + INSERT SELECT + RENAME, or rewrite at import time.
- `otel_logs` is the table that powers the frontend Traces/Sessions page, not the `traces` table. The `traces` table is used by the GraphQL dashboard (overview, trends).
- `owner_org_id` was added in migration 0019. Archives from before that migration have NULL values that make records invisible to org-scoped API queries.
- ClickHouse block-level insert deduplication silently drops INSERT...SELECT even with different column values. Use `insert_deduplicate=0` or the temp table + RENAME approach to work around it.

## Checklist

### DIFF
--- a/observal_cli/cmd_migrate.py
+++ b/observal_cli/cmd_migrate.py
@@ -112,6 +112,9 @@ CLICKHOUSE_TABLES: list[TableCfg] = [
     {"name": "audit_log", "engine": "mergetree", "time_col": "timestamp", "fk_cols": ["actor_id"]},
     {"name": "mcp_tool_calls", "engine": "mergetree", "time_col": "timestamp", "fk_cols": ["mcp_server_id", "user_id"]},
     {"name": "agent_interactions", "engine": "mergetree", "time_col": "timestamp", "fk_cols": ["agent_id", "user_id"]},
+    {"name": "otel_logs", "engine": "mergetree", "time_col": "Timestamp", "fk_cols": []},
+    {"name": "security_events", "engine": "mergetree", "time_col": "timestamp", "fk_cols": []},
+    {"name": "webhook_deliveries", "engine": "mergetree", "time_col": "timestamp", "fk_cols": []},
 ]
 
 FK_PG_TABLE_MAP: dict[str, str] = {
@@ -411,7 +414,14 @@ async def _insert_table(
             row = json.loads(line)
 
             if columns is None:
-                columns = list(row.keys())
+                target_columns = set(col_types.keys())
+                columns = [k for k in row if k in target_columns]
+                skipped_cols = set(row) - target_columns
+                if skipped_cols:
+                    rprint(
+                        f"[dim]  {jsonl_path.stem}: skipping archive columns not in target: "
+                        f"{', '.join(sorted(skipped_cols))}[/dim]"
+                    )
 
             batch.append(row)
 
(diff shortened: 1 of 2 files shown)


Haz3-jolt:
Adversarial review. No blockers, three nits.

The otel_logs entry uses time_col: Timestamp (capital T) while every other entry uses lowercase timestamp. If the actual ClickHouse DDL column is capitalized this is correct, but it is the only entry that differs and there is no comment explaining why. Worth a one-liner comment so future readers do not assume it is a typo.

Column intersection is computed from the first JSONL row only. If the archive has sparse rows where later rows contain columns not present in the first row that do exist in the target schema, those columns are silently dropped for all subsequent rows. Unlikely with well-formed exports but a latent assumption.

_rewrite_project_id materializes a full Python list with pa.array([target_project_id] * len(table)). For large tables this creates N Python string objects in memory before converting to a PyArrow array. Not a problem at typical scale but PyArrow can create constant columns without the intermediate list allocation.

Everything else (column filtering logic, schema mismatch downgrade to warning, resume state validation, owner_org_id backfill, missing tables addition) is correct. Clean to merge.

PR #741, opened 2026-05-03
fix(review): flush version changes before updating latest_version_id

### LINKED ISSUE #740: bug: Approve components throws 500 error in CLI and UI
### Checked for duplicates?

1. login as any user, create a component.
2. go to reviews -> click on approve

### Expected behaviour

the component should have been approved successfully

_No response_

ShaanNarendran:
## Purpose / Description

The component approve endpoint (`POST /api/v1/review/{listing_id}/approve`) throws a 500 error (SQLAlchemy `CircularDependencyError`) when approving components that have versioning enabled. This is because the endpoint modifies both the version's status and the listing's `latest_version_id` FK in a single commit without flushing intermediate changes.

The bidirectional FK relationship (listing → version via `latest_version_id`, version → listing via `listing_id`) requires an explicit `db.flush()` between modifications to avoid the circular dependency during commit.

## Fixes
* Fixes #740

## Approach

Added `await db.flush()` after updating the pending version's fields and before setting `listing.latest_version_id`, matching the pattern already used in `approve_agent()` (which has the same comment: "Flush version changes first to avoid CircularDependencyError").

## How Has This Been Tested?

- Added `test_approves_pending_version_and_updates_latest` to `tests/test_review_queue.py` which verifies:
  - The pending version's status is set to approved
  - `listing.latest_version_id` is updated to the pending version
  - `db.flush()` is called before `db.commit()`
- All 22 tests in `test_review_queue.py` pass

In async SQLAlchemy, when two models have bidirectional foreign keys (A.b_id → B, B.a_id → A), modifying both sides in a single `session.commit()` can trigger a `CircularDependencyError`. The fix is to `await session.flush()` after modifying one side before modifying the other, so SQLAlchemy can resolve the dependency order.

## Checklist

### DIFF
--- a/observal-server/api/routes/review.py
+++ b/observal-server/api/routes/review.py
@@ -548,6 +548,8 @@ async def approve(
         pending_ver.rejection_reason = None
         pending_ver.reviewed_by = current_user.id
         pending_ver.reviewed_at = datetime.now(UTC)
+        # Flush version changes first to avoid CircularDependencyError
+        await db.flush()
         # Update latest_version_id to point to newly approved version
         listing.latest_version_id = pending_ver.id
     else:
--- a/tests/test_review_queue.py
+++ b/tests/test_review_queue.py
@@ -409,6 +409,31 @@ class TestApprove:
         assert "id" in data
 
     @pytest.mark.asyncio
+    async def test_approves_pending_version_and_updates_latest(self):
+        """When a listing has a pending version, approve targets the version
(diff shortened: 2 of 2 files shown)


ShaanNarendran:
commit which had to be changed:
a4aef71 — "fix(review): version-aware component review queue, detail, and actions" by Hari Srinivasan (May 2, 23:35 UTC)

PR #744, opened 2026-05-03
feat(eval): sessions by agent for claude

### LINKED ISSUE #743: [FEATURE] Session matching for evals for non-Kiro
### Is your feature request related to a problem? Please describe.

Claude Code does not populate the agent_name field in telemetry events, making it impossible to query which sessions were run with a specific agent.

Currently:

Kiro CLI sets LogAttributes['agent_name'] → sessions can be queried by agent
Claude Code leaves LogAttributes['agent_name'] empty → sessions cannot be attributed to specific agents
Impact: The Observal eval system cannot find evaluated sessions for agents run via Claude Code. The session selection dropdown shows "No evaluated sessions" even when sessions exist, forcing users to use Kiro instead.

### Describe the solution you'd like

Add agent_name to the telemetry payload when Claude Code sends hook events (pretooluse, posttooluse, etc.). The field should populate from the active agent's manifest/config, matching how Kiro already implements this.

Expected: All hook events should include LogAttributes['agent_name'] matching the agent identifier being used.

### Describe alternatives you've considered

_No response_

### Additional context

The Observal server query that finds evaluated sessions filters by agent_name:

WHERE LogAttributes['agent_name'] = {agent_name}
This returns 0 rows for Claude Code sessions. Kiro already populates this field correctly - Claude Code should follow the same pattern for observability parity between IDEs.

VishnuM049:
## Purpose / Description
Needed to make sure claude sessions were taken into account for agent eval

## Fixes
* Fixes #743

## Approach
uses the agent specific .claude files, adding am agent-name field to check against for each session

## How Has This Been Tested?
full flow checked to see results in frontend with multiple concurrent agents

_Describe the research stage_

## Checklist

### DIFF
--- a/observal-server/api/routes/sessions.py
+++ b/observal-server/api/routes/sessions.py
@@ -869,6 +869,8 @@ async def get_session_efficiency(session_id: str, current_user: User = Depends(r
             "ServiceName AS service_name "
             "FROM otel_logs "
             "WHERE TraceId = {sid:String} "
+            "AND (LogAttributes['user.id'] = {uid:String} "
+            "     OR LogAttributes['user.id'] = {uemail:String}) "
             "ORDER BY Timestamp ASC",
             params,
         )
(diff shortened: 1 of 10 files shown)


Haz3-jolt (commented):
Reviewed the changes. A few things to address before merge:

**Must fix**

`bind_session_agent` (sessions.py) has no ownership check. Any authenticated user who knows or guesses a session_id can overwrite another user's agent attribution in Redis. Every other session endpoint gates on `LogAttributes['user.id']` for non-admins. This needs the same pattern before the Redis write.

`_resolve_hook_paths` in cmd_pull.py switched from quote-bounded replacement (`content.replace(f'"{name}"', ...)`) to bare substring replacement (`content.replace(name, ...)`). This will match the script name anywhere in the file content, not just inside command values. If an agent's prompt text or a comment mentions "observal-hook.sh", it gets corrupted. Anchor the replacement to the command context, e.g. regex matching `"observal-hook.sh..."` with optional trailing args inside the quotes.

**Should fix**

`$_py` in observal-hook.sh (line 43) is unquoted. If the Python executable path contains spaces (e.g. a virtualenv under a directory with spaces), word splitting causes silent fallthrough to the `|| echo "$payload"` branch and agent attribution is silently dropped. Quote it as `"$_py"`.

`_claude_code_hooks_frontmatter_lines` in agent_config_generator.py accepts a raw `agent_name` param and interpolates it directly into a shell command string. The function itself has no sanitization guard. It relies entirely on the caller having already sanitized. If any future code path calls this without sanitizing, you get command injection via YAML frontmatter. A one-line `re.sub(r"[^a-zA-Z0-9_\-.]", "", agent_name)` at the top of the function would make it safe in isolation.

The TraceId fallback query in `get_session_efficiency` (sessions.py, the new block) does not re-run the ownership check. The ownership query only validates `LogAttributes['session.id']`. When the fallback fires using TraceId, non-admin users can potentially read events from sessions they do not own.

**Minor/optional**

The auto-detect fallback in cmd_hook.py (`len(agents) == 1`) silently writes `OBSERVAL_AGENT_NAME` to the env config without telling the user. A one-line info print would avoid confusion.

`session_id` used as a Redis key suffix in telemetry.py is externally supplied and has no format validation. Not exploitable in practice but a character/length check would be defensive.

Haz3-jolt (approved):

PR #751, opened 2026-05-05
fix: replace dangerous alembic stamp-head fallback with smart migration script

SrihariLegend:
The entrypoint.sh previously ran 'alembic stamp head' when migrations failed, silently marking all migrations as applied without running them. This left the database with missing columns (e.g. gaming_flags on agent_versions) when users started the stack without a .env file first.

Replace with docker/migrate.py that detects the database state:
- Empty DB: create_all + stamp head (fresh install)
- Managed DB: alembic upgrade head, fail hard on error (never stamp)
- Unmanaged DB: create_all + sync missing columns via inspect() + stamp

The sync_missing_columns() function uses SQLAlchemy inspect() to diff model-defined columns against the live DB and issues ALTER TABLE ADD COLUMN IF NOT EXISTS for any gaps before stamping.

Also removes the _ensure_columns hack from main.py since schema reconciliation is now fully handled by the init container.

## Purpose / Description
The Docker init container's entrypoint.sh uses alembic stamp head as a fallback when migrations fail. This silently
marks all migrations as applied without actually running them, leaving the database with missing columns. This causes
500 errors across all API endpoints (agents, review, evals, insights).

The issue is triggered when a user starts the stack without a .env file first (or with incorrect DB credentials), then
adds the correct .env and rebuilds. The Base.metadata.create_all creates tables from the current models, Alembic
migrations fail (because they assume incremental state), and the fallback stamp head marks everything as done, but
columns added by later migrations (e.g., gaming_flags on agent_versions from migration 0028) are never created.

## Fixes
Fixes the 500 errors caused by missing agent_versions.gaming_flags column after rebuild and the dangerous alembic stamp head fallback that silently hides migration failures.

## Approach
Replace the fragile bash migration logic in entrypoint.sh with a Python script (docker/migrate.py) that detects the
database state and applies the correct strategy:

1. Empty DB (no tables): create_all + stamp head -> fresh install, models are source of truth.
2. Managed DB (has alembic_version): alembic upgrade head -> fail hard on error, never stamp.
3. Unmanaged DB (tables exist, no alembic_version): create_all for missing tables + sync_missing_columns() using
SQLAlchemy inspect() to diff model columns vs live DB and issue ALTER TABLE ADD COLUMN IF NOT EXISTS for any gaps, then
stamp head.

Also removes the _ensure_columns hack from main.py since schema reconciliation is now fully handled by the init
container.

## How Has This Been Tested?

Unit tests (31 tests in tests/test_docker_migrate.py):

bash
cd observal-server && uv run --with pytest --with pytest-asyncio pytest ../tests/test_docker_migrate.py -v

Tests cover:
- detect_state_sync() — all three states + edge cases (NULL count, DB errors propagating, single-table detection)
- sync_missing_columns() — no-op when schema matches, adds nullable columns, handles server defaults, NOT NULL fallback
to nullable, multiple missing columns, skips nonexistent tables, warns on unrecoverable failures
- run_alembic() — exit code passthrough, subprocess errors propagate
- main() flow — all three paths succeed, all failure paths exit non-zero
- Adversarial inputs — SQL injection in column names (quoted), empty metadata, 500-column stress test, inspect()
failures propagate

Full test suite:

bash
cd observal-server && uv run --with pytest --with pytest-asyncio --with pyyaml --with typer --with rich --with hypothesis pytest ../tests/ -q --ignore=../tests/test_migrate_telemetry.py

Result: 2035 passed, 3 pre-existing failures (unrelated), 0 new failures.

Test configuration: Python 3.12, pytest 9.0.3, pytest-asyncio 1.3.0, all tests mock external services (no Docker needed)

_Describe the research stage_

## Checklist

### DIFF
--- a/docker/Dockerfile.api
+++ b/docker/Dockerfile.api
@@ -25,6 +25,7 @@ COPY --from=builder --chown=appuser:appgroup /app/.venv /app/.venv
 COPY --chown=appuser:appgroup observal-server/ .
 COPY --chown=appuser:appgroup ee/ ./ee/
 COPY --chown=appuser:appgroup docker/entrypoint.sh /app/entrypoint.sh
+COPY --chown=appuser:appgroup docker/migrate.py /app/migrate.py
 RUN sed -i 's/\r$//' /app/entrypoint.sh
 
 USER appuser
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -1,25 +1,8 @@
 #!/bin/bash
 set -e
 
-echo "Ensuring base schema exists..."
-/app/.venv/bin/python -c "
-import asyncio
-from database import engine
-from models import Base
-
-async def init():
-    async with engine.begin() as conn:
-        await conn.run_sync(Base.metadata.create_all)
-    await engine.dispose()
-
-asyncio.run(init())
-"
-
-echo "Running database migrations..."
-/app/.venv/bin/python -m alembic upgrade head || {
-    echo "Migrations failed — stamping head and retrying..."
-    /app/.venv/bin/python -m alembic stamp head
-}
+echo "Running database migration..."
+/app/.venv/bin/python /app/migrate.py
 
 echo "Ensuring ClickHouse database exists..."
 # Parse credentials from CLICKHOUSE_URL using Python to handle special chars
--- a/docker/migrate.py
+++ b/docker/migrate.py
@@ -1,0 +1,175 @@
+"""Smart database migration script for Observal.
+
+Detects the current database state and applies the correct migration strategy:
+- Empty DB: create_all + stamp head (fresh install)
+- Managed DB (has alembic_version): alembic upgrade head (fail hard on error)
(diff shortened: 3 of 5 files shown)


SrihariLegend:
[40](https://github.com/BlazeUp-AI/Observal/actions/runs/25394758449/job/74478404659?pr=751#step:7:341)
FAILED ../tests/test_draft_workflow.py::TestDraftSubmit::test_transitions_to_pending - TypeError: expected string or bytes-like object, got 'MagicMock'
FAILED ../tests/test_draft_workflow.py::TestDraftSubmit::test_submit_response_includes_status - TypeError: expected string or bytes-like object, got 'MagicMock'
FAILED ../tests/test_worker_phase5.py::TestWorkerSettings::test_job_timeout - AssertionError: assert 600 == 300


All 3 tests that fail does not have any relation to this pull request.

Haz3-jolt:
We are squashing to a new baseline. No point for backward compat right now. Will just introduce more regressions.

SrihariLegend:
close the PR

Issue #672, opened 2026-05-01
[BUG] Shallow copy alembic import issue in migrations


Lokesh7025:
### Checked for duplicates?

- [X] This issue is not a duplicate

### What are the steps to reproduce this bug?

1. Run `observal migrate` or trigger the alembic migration path
2. The migration uses a shallow copy of alembic imports or config
3. This causes import errors or incorrect migration behavior when crossing versions

### Expected behaviour

Alembic imports should be handled with proper deep copies or direct references so that cross-version migrations work correctly without import errors.

ShaanNarendran:
Should be fixed by #753

PR #753, opened 2026-05-05
refactor(migrations): squash 28 migrations into single baseline

Haz3-jolt:
## Purpose / Description
Replace 28 incremental alembic migrations with a single no-op baseline revision. We have no external users — the hosted database already has every table and column. The migration chain adds complexity without value and was causing issues (see #751) when the entrypoint tried to handle edge cases around partially-migrated databases.

## Fixes
* Supersedes #751 — replaces the "smart migration script" approach with a simpler squash
* #672 

## Approach
- Delete all 28 migration files (0001–0028)
- Add a single `0001_initial_schema.py` with `upgrade = pass` — it exists solely as the alembic baseline marker
- Keep the entrypoint simple: `create_all` for tables, then `alembic upgrade head` (with stamp fallback for fresh DBs)
- No `docker/migrate.py`, no column-sync logic, no state detection

For the hosted deploy: the existing `alembic_version` row (currently `0028`) won't match the new `0001` revision. The deploy will fail on `alembic upgrade head`, trigger the stamp fallback, and set the version to `0001`. Since all tables already exist, this is safe — it's just resetting bookkeeping.

## How Has This Been Tested?
- Verified `alembic.ini` and `env.py` are unchanged (migration infrastructure intact)
- The no-op migration is valid Python that alembic can parse
- Fresh DB path: `create_all` makes tables → `upgrade head` fails (tables exist) → stamp sets version to 0001 ✓
- Existing DB path: `upgrade head` fails (0028 not in chain) → stamp resets to 0001 ✓
- Future migrations: will use `down_revision = "0001"` and run normally via `upgrade head` ✓

## Checklist


ShaanNarendran:
if it was tested, this is approved from me. I think our schemas have been finalized so this is a needed change.

ShaanNarendran (approved):

SrihariLegend:
make sure that the entrypoint makes create_all safe for existing databases and the stamp fallback is narrowed to only trigger on the missing revision case, not on all upgrade head failures. If so then all good.

SrihariLegend:
does create_all succeed whether the tables are brand new or already present without errors?

Haz3-jolt:
Yes to both, this is just how SQLAlchemy works:

1) create_all() uses checkfirst=True by default, it checks has_table() before emitting any DDL. It's been idempotent since forever. It won't error on existing tables. This isn't something we need to "make safe,", it's the documented default behavior.

2) On the stamp fallback, the migration is literally pass. There is no SQL in it. The only way upgrade head fails is a revision chain mismatch, which is exactly and exclusively what stamp head exists to fix. There's nothing to "narrow" here, there's one failure mode and we handle it.

Both of these are covered in the PR description under "How Has This Been Tested."

PR #760, opened 2026-05-07
feat(ee): merge observal-insights into ee/ for enterprise distribution

ShaanNarendran:
## Purpose / Description

Merges the standalone `observal-insights` package (from `BlazeUp-AI/observal-insights`) into the main Observal repo under `ee/observal_insights/`. This makes the Agent Insights engine part of the enterprise edition, covered under the EE license, and removes the need for a separate package install.

Enterprise gating is now based on `DEPLOYMENT_MODE=enterprise` environment variable rather than package availability, returning HTTP 402 for non-enterprise users.

## Fixes
* Consolidates the insights engine into the monorepo for simpler distribution and licensing

## Approach

1. **Ported all 18 modules** from `observal-insights/src/observal_insights/` into `ee/observal_insights/`
2. **Converted imports** from absolute (`from observal_insights.X`) to relative (`from .X`)
3. **Updated enterprise gating** in `observal-server/services/insights/__init__.py` to check `DEPLOYMENT_MODE == "enterprise"` instead of checking for pip package availability
4. **Simplified Docker setup** — removed the optional pip install block from `Dockerfile.api` since code is now bundled
5. **Updated `docker-compose.enterprise.yml`** — just sets the env var, no separate image needed
6. **Updated Makefile** enterprise detection to check for `ee/observal_insights/__init__.py`
7. **Added ruff per-file-ignores** for the ported code to preserve original style

The dependency injection pattern (`_deps.py` / `configure()`) is preserved — the insights package never imports from the server directly.

## How Has This Been Tested?

- Full test suite passes (2005 tests collected, all pass)
- `ruff check` passes clean on entire repo
- Import boundary enforcement verified (core cannot import from `ee/`)

- Using ruff's `TID251` (flake8-tidy-imports banned-api) rule to enforce the ee/ import boundary at lint time
- Dependency injection via a `configure()` + module-level container avoids circular imports between ee/ and server code

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -201,7 +201,7 @@ jobs:
       - name: Check ee/ import boundary
         run: |
           echo "Checking that core does not import from ee/..."
-          VIOLATIONS=$(grep -rn "from ee\b\|import ee\b" observal-server/ observal_cli/ --include="*.py" | grep -v "main.py" || true)
+          VIOLATIONS=$(grep -rn "from ee\b\|import ee\b" observal-server/ observal_cli/ --include="*.py" | grep -v "main.py" | grep -v "services/insights/__init__.py" || true)
           if [ -n "$VIOLATIONS" ]; then
             echo "::error::Import boundary violation — core must not import from ee/"
             echo "$VIOLATIONS"
--- a/Makefile
+++ b/Makefile
@@ -39,11 +39,11 @@ hooks:  ## Install pre-commit hooks
 
 # ── Docker ───────────────────────────────────────────────
 
-# Auto-detect enterprise mode: if observal-insights/ exists, use enterprise override
+# Auto-detect enterprise mode: if ee/observal_insights/ exists, use enterprise override
 COMPOSE_FILES := -f docker-compose.yml
-ifneq (,$(wildcard observal-insights/pyproject.toml))
+ifneq (,$(wildcard ee/observal_insights/__init__.py))
   COMPOSE_FILES += -f docker-compose.enterprise.yml
-  $(info [enterprise mode] observal-insights/ detected)
+  $(info [enterprise mode] ee/observal_insights/ detected)
 endif
 
 up:  ## Start Docker stack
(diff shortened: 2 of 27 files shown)


ShaanNarendran:
## Follow-up fixes pushed

### 1. `fix(clickhouse): add TTL retention for system logs to prevent OOM`

During testing, ClickHouse was hitting OOM (memory limit exceeded at 1.6G) which caused **all telemetry inserts to fail silently** — hooks fired, the API returned 200, but data never reached ClickHouse.

**Root cause:** ClickHouse's internal system tables (`trace_log`, `text_log`, `metric_log`, `asynchronous_metric_log`, `part_log`) were accumulating unbounded — ~410 MB on disk, 75M+ rows, far more in memory. With the container capped at 2G (1.6G usable at 80% ratio), there was no headroom left for actual queries or inserts.

**Fix:** Added TTL retention (1–3 days) to all system log tables in `docker/clickhouse/config.d/memory.xml` so they auto-prune and never accumulate enough to threaten the container's memory budget.

### 2. `fix(insights): align session_cache and facets with actual DB schema`

The ported `session_cache.py` expected a model with `period_start`, `period_end`, and `session_metas` columns (one row per time range). The actual `InsightSessionMeta` model stores **one row per session** with `(agent_id, session_id, meta)`. Similarly, `facets.py` referenced `created_at`/`updated_at` columns that don't exist — the real model uses `extracted_at`.

This mismatch caused `'InsightSessionMeta' has no attribute 'period_start'` errors, forcing fallback to ClickHouse (which was OOM), resulting in 0 sessions analyzed.

**Fix:** Rewrote `load_cached_metas` and `store_cached_metas` to work with per-session rows, and updated `store_facets` to use the correct column names.

ShaanNarendran:
### 3. `fix(docker): pin pnpm@10 to fix ERR_PNPM_IGNORED_BUILDS in CI`

The `docker-build` CI check was still failing after the initial fix because `node:24-alpine` ships **pnpm 11** via corepack. pnpm 11 changed behavior: it now exits with code 1 on `ERR_PNPM_IGNORED_BUILDS` regardless of `ignoredBuiltDependencies` or `onlyBuiltDependencies` config — it's a hard failure, not a warning.

**Fix:** Added `"packageManager": "pnpm@10.33.0"` to `web/package.json` so corepack pins to pnpm 10 (which respects the workspace config), and also copy `pnpm-workspace.yaml` into the Docker deps stage so pnpm sees the `ignoredBuiltDependencies` for `sharp` and `unrs-resolver`.

Haz3-jolt (approved):

PR #787, opened 2026-05-09
fix: add version tables to INSERT_ORDER and fix related import issues

### LINKED ISSUE #786: bug: observal migrate import fails due to missing version tables in INSERT_ORDER
### Checked for duplicates?

## Steps to reproduce

1. Have a source database with component versions (`mcp_versions`, `skill_versions`, `hook_versions`, `prompt_versions`, `sandbox_versions`, `agent_versions`) and listings that reference them via `latest_version_id`
2. Run `observal migrate export --db-url <source-db>` - exports all tables including version tables
3. Run `observal migrate import --db-url <target-db> --archive <archive>` on a clean target database
4. Observe massive FK violations - all listings, agents, and downstream tables fail to import

### Expected behaviour

All rows should import successfully. Instead, only 20 out of 417 rows imported and 397 were skipped due to FK violations.

**Root cause:** `INSERT_ORDER` in `observal_cli/cmd_migrate.py` is missing the `*_versions` tables (`mcp_versions`, `skill_versions`, `hook_versions`, `prompt_versions`, `sandbox_versions`, `agent_versions`). These tables are exported in the archive but never imported. Since listings and agents have a `latest_version_id` FK pointing to version tables, they all fail with `*_latest_version_id_fkey` violations, which cascades to every downstream table (downloads, eval_runs, scorecards, scorecard_dimensions, agent_components, etc.).

Additionally, the `ON CONFLICT ("id") DO NOTHING` clause in `_build_insert` doesn't handle unique constraint violations on non-PK columns (e.g. `organizations.slug`, `users.email`). If the target DB has bootstrap/demo data with matching slugs or emails but different UUIDs, the import crashes with `UniqueViolationError` instead of gracefully skipping.

naraen-ram:
## Purpose / Description
Fix `observal migrate import` failing with FK violations due to missing version tables in INSERT_ORDER, plus several related import robustness issues.

## Fixes
* Fixes #786

## Approach
- Added 6 version tables (`mcp_versions`, `skill_versions`, `hook_versions`, `prompt_versions`, `sandbox_versions`, `agent_versions`) to `INSERT_ORDER`
- Disabled FK constraints during import using `SET session_replication_role = 'replica'` to handle circular dependencies between listings and their version tables
- Catch `UniqueViolationError` gracefully (skip with warning instead of crash)
- Serialize JSON/JSONB values properly for asyncpg (dict → JSON string)
- Handle NOT NULL JSON columns with defaults (`models_by_ide`)
- Skip tables not present in older archives gracefully

## How Has This Been Tested?
- Exported from dev.observal.io (39 tables, 469 rows including version tables)
- Imported into a fresh local instance — 465 inserted, 4 skipped (expected demo account conflicts)
- Verified agents, components, eval runs, and scorecards all visible in the UI
- Tested all 6 CLI commands end-to-end: export, validate, import, export-telemetry, import-telemetry, validate-telemetry
- Confirmed idempotent re-import (all rows skipped on second run)

## Execution Example

```
# Phase 1: Shallow copy (PostgreSQL)
observal migrate export --db-url <source-postgres-url> --output ./migration/dev-export.tar.gz
observal migrate validate --archive ./migration/dev-export.tar.gz
observal migrate import --db-url <target-postgres-url> --archive ./migration/dev-export.tar.gz

# Phase 2: Deep copy (ClickHouse telemetry)
observal migrate export-telemetry --clickhouse-url <source-clickhouse-url> --manifest ./migration/dev-export.manifest.json --output-dir ./migration/telemetry
observal migrate import-telemetry --clickhouse-url <target-clickhouse-url> --input-dir ./migration/telemetry --project-id <target-org-id>
observal migrate validate-telemetry --input-dir ./migration/telemetry --clickhouse-url <target-clickhouse-url>
```

## Known Gaps

**Org UUID remapping (PostgreSQL):**
When source and target have different org UUIDs (which they always will since the bootstrap creates a new org), all imported data references the source org ID. There's no `--org-id` flag on `migrate import` to auto-remap. Currently requires manual SQL UPDATEs on `agents.owner_org_id`, `users.org_id`, and all `*_listings.owner_org_id` tables after import. The `import-telemetry` command has `--project-id` for ClickHouse but there's no PostgreSQL equivalent.

**Traces not visible in "My Traces" UI after deep copy:**
The "My Traces" page displays reconciled sessions (from `observal reconcile`), not raw ClickHouse traces. Raw telemetry is imported correctly and queryable via GraphQL/API, but the session-level UI requires reconciliation data which lives in a different data path. Additionally, the ClickHouse `traces` table uses `project_id` as part of the primary key — if not rewritten during import (via `--project-id`), traces won't match the local org and won't appear in any org-scoped queries. The `project_id` cannot be mutated after insert since it's in the sort key.

## Checklist
- [ ] UI changes: include screenshots of all affected screens (N/A - backend only)

### DIFF
--- a/observal_cli/cmd_migrate.py
+++ b/observal_cli/cmd_migrate.py
@@ -44,12 +44,23 @@ INSERT_ORDER: list[str] = [
     # Tier 1.5 — FK to users
     "component_bundles",
     # Tier 2 — FK to orgs + users + component_bundles
+    # NOTE: listings/agents have a circular FK with their version tables:
+    #   *_listings.latest_version_id → *_versions.id (nullable, use_alter)
+    #   *_versions.listing_id → *_listings.id (NOT NULL)
+    # We break the cycle by nulling latest_version_id during insert (see DEFERRED_VERSION_BACKFILL).
     "mcp_listings",
     "skill_listings",
     "hook_listings",
     "prompt_listings",
     "sandbox_listings",
     "agents",
+    # Tier 2.5 — FK to listings/agents + users (version tables)
+    "mcp_versions",
+    "skill_versions",
+    "hook_versions",
+    "prompt_versions",
+    "sandbox_versions",
+    "agent_versions",
     # Tier 3 — FK to listings/users
     "mcp_validation_results",
     "mcp_downloads",
@@ -59,14 +70,14 @@ INSERT_ORDER: list[str] = [
     "sandbox_downloads",
     "submissions",
     "alert_rules",
-    # Tier 4 — FK to agents
+    # Tier 4 — FK to agents/agent_versions
     "agent_goal_templates",
     "agent_download_records",
     "component_download_records",
     "dimension_weights",
     # Tier 5 — FK to agent_goal_templates
     "agent_goal_sections",
-    # Tier 6 — FK to agents (polymorphic component_id)
+    # Tier 6 — FK to agent_versions (polymorphic component_id)
     "agent_components",
     # Tier 7 — FK to users (polymorphic listing_id)
     "feedback",
@@ -81,13 +92,40 @@ INSERT_ORDER: list[str] = [
     "trace_penalties",
 ]
 
+# Tables with a circular latest_version_id FK that must be nulled during insert
+# and backfilled after version tables are imported.
+# Format: (table, version_table, fk_column)
+DEFERRED_VERSION_BACKFILL: list[tuple[str, str, str]] = [
+    ("mcp_listings", "mcp_versions", "latest_version_id"),
+    ("skill_listings", "skill_versions", "latest_version_id"),
+    ("hook_listings", "hook_versions", "latest_version_id"),
+    ("prompt_listings", "prompt_versions", "latest_version_id"),
+    ("sandbox_listings", "sandbox_versions", "latest_version_id"),
+    ("agents", "agent_versions", "latest_version_id"),
+]
+
 JSONB_COLUMNS: dict[str, list[str]] = {
     "agents": ["model_config_json", "external_mcps", "supported_ides"],
+    "agent_versions": [
+        "model_config_json",
+        "external_mcps",
+        "supported_ides",
+        "required_ide_features",
+        "inferred_supported_ides",
+        "ide_configs",
+        "gaming_flags",
+        "models_by_ide",
+    ],
     "mcp_listings": ["tools_schema", "environment_variables", "supported_ides"],
+    "mcp_versions": ["tools_schema", "environment_variables", "supported_ides", "args", "headers", "auto_approve"],
     "skill_listings": ["supported_ides", "target_agents", "triggers", "mcp_server_config", "activation_keywords"],
+    "skill_versions": ["supported_ides", "target_agents", "triggers", "mcp_server_config", "activation_keywords"],
     "hook_listings": ["supported_ides", "handler_config", "input_schema", "output_schema"],
+    "hook_versions": ["supported_ides", "handler_config", "input_schema", "output_schema"],
     "prompt_listings": ["variables", "model_hints", "tags", "supported_ides"],
+    "prompt_versions": ["variables", "model_hints", "tags", "supported_ides"],
     "sandbox_listings": ["resource_limits", "allowed_mounts", "env_vars", "supported_ides"],
+    "sandbox_versions": ["resource_limits", "allowed_mounts", "env_vars", "supported_ides"],
     "scorecards": ["raw_output", "dimension_scores", "scoring_recommendations", "dimensions_skipped", "warnings"],
     "agent_components": ["config_override"],
     "exporter_configs": ["config"],
@@ -334,7 +372,16 @@ def _coerce_value(value: object, pg_type: str) -> object:
         return int(value)
     if pg_type in ("float4", "float8", "numeric") and isinstance(value, (int, float)):
         return float(value)
+    # asyncpg requires JSON/JSONB values as serialized strings
+    if pg_type in ("json", "jsonb") and not isinstance(value, str):
+        return json.dumps(value)
     return value
+
+
+# Columns that are NOT NULL with JSON defaults — used when archive has NULL for these.
+_NOTNULL_JSON_DEFAULTS: dict[str, dict[str, str]] = {
+    "agent_versions": {"models_by_ide": "{}"},
+}
 
 
 def _build_insert(table: str, columns: list[str], col_types: dict[str, str]) -> str:
@@ -376,6 +423,12 @@ async def _flush_batch(
     skipped = 0
 
     for row in batch:
+        # Apply NOT NULL JSON defaults for columns that are NULL in the archive
+        table_defaults = _NOTNULL_JSON_DEFAULTS.get(table, {})
+        for col, default_val in table_defaults.items():
+            if col in columns and row.get(col) is None:
+                row[col] = default_val  # Already a JSON string
+
         values = [_coerce_value(row.get(col), col_types.get(col, "")) for col in columns]
         try:
             status = await conn.execute(query, *values)
@@ -389,6 +442,10 @@ async def _flush_batch(
             row_id = row.get("id", "unknown")
             rprint(f"[yellow]  FK violation in {table}, row {row_id}: {e.constraint_name}[/yellow]")
             skipped += 1
+        except asyncpg.UniqueViolationError as e:
+            row_id = row.get("id", "unknown")
+            rprint(f"[yellow]  Unique conflict in {table}, row {row_id}: {e.constraint_name}[/yellow]")
+            skipped += 1
 
     return inserted, skipped
 
@@ -398,6 +455,8 @@ async def _insert_table(
     table: str,
     jsonl_path: Path,
     col_types: dict[str, str],
+    org_rewrite_map: dict[str, str] | None = None,
+    org_columns: set[str] | None = None,
 ) -> tuple[int, int]:
     """Insert rows from a JSONL file into a table. Returns (inserted, skipped)."""
     inserted = 0
@@ -405,6 +464,9 @@ async def _insert_table(
     batch: list[dict] = []
     columns = sorted(col_types.keys())
     logged_skipped = False
+
+    # Determine which columns in this table need org rewriting
+    rewrite_cols = (org_columns & set(columns)) if org_rewrite_map and org_columns else set()
 
     with open(jsonl_path, encoding="utf-8") as f:
         for line in f:
@@ -421,6 +483,13 @@ async def _insert_table(
                         f"{', '.join(sorted(skipped_cols))}[/dim]"
                     )
                     logged_skipped = True
+
+            # Rewrite org IDs if normalization is active
+            if rewrite_cols and org_rewrite_map:
+                for col in rewrite_cols:
+                    val = row.get(col)
+                    if val and val in org_rewrite_map:
+                        row[col] = org_rewrite_map[val]
 
             batch.append(row)
 
@@ -663,7 +732,7 @@ def _is_empty_parquet(path: Path) -> bool:
         return True
 
 
-async def _import_archive(db_url: str, archive_path: Path) -> ImportResult:
+async def _import_archive(db_url: str, archive_path: Path, normalize_org_id: str | None = None) -> ImportResult:
     """Import a migration archive into the target database."""
     t0 = time.monotonic()
     warnings: list[str] = []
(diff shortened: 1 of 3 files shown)


Haz3-jolt:
Hi! We recently added a CLA requirement to the repo. Could you rebase your branch on main and force push? That will trigger the CLA bot to comment with the signing link. Thanks!

naraen-ram:
## Org ID Normalization (`--org-id`)

When importing into a target that has a different org UUID than the source (which is always the case since bootstrap creates a new org), all `org_id` and `owner_org_id` references in the imported data will point to a non-existent org. This causes agents and components to be invisible in the UI.

The `--org-id` flag on `migrate import` fixes this by auto-detecting source org UUIDs from the archive and rewriting all org references to the target org during insert. This mirrors how `--project-id` works on `import-telemetry` for ClickHouse.

### Usage

```bash
# Get your local org ID
docker exec docker-observal-db-1 psql -U postgres -d observal -c "SELECT id FROM organizations LIMIT 1;" -t

# Import with org normalization
observal migrate import \
  --db-url <target-postgres-url> \
  --archive ./migration/dev-export.tar.gz \
  --org-id <target-org-id>

**Review by Haz3-jolt — APPROVED** [review]

File: observal_cli/cmd_migrate.py [hunk 1]
```diff
@@ -81,13 +92,40 @@
...
+# and backfilled after version tables are imported.
+# Format: (table, version_table, fk_column)
+DEFERRED_VERSION_BACKFILL: list[tuple[str, str, str]] = [
+    ("mcp_listings", "mcp_versions", "latest_version_id"),
+    ("skill_listings", "skill_versions", "latest_version_id"),
+    ("hook_listings", "hook_versions", "latest_version_id"),
+    ("prompt_listings", "prompt_versions", "latest_version_id"),
+    ("sandbox_listings", "sandbox_versions", "latest_version_id"),
+    ("agents", "agent_versions", "latest_version_id"),
+]
```
Apoorvgarg-creator, line 105: This constant and the `deferred_tables` map built from it at lines 802 to 805 aren't read anywhere downstream. The leading comment promises we 'break the cycle by nulling latest_version_id during insert (see DEFERRED_VERSION_BACKFILL)', but the actual cycle break is the `session_replication_role = 'replica'` switch a few lines below.

So either this is leftover scaffolding from an earlier approach (delete it), or the intended 'null then backfill' flow was never finished. The two strategies have meaningfully different blast radii (see the comment on line 828), so worth deciding which one we're keeping and removing the other before merge. Carrying both makes the next reader think one is wired up when it isn't.

observal_cli/cmd_migrate.py
```diff
@@ -334,9 +372,18 @@ def _coerce_value(value: object, pg_type: str) -> object:
         return int(value)
     if pg_type in ("float4", "float8", "numeric") and isinstance(value, (int, float)):
         return float(value)
+    # asyncpg requires JSON/JSONB values as serialized strings
+    if pg_type in ("json", "jsonb") and not isinstance(value, str):
+        return json.dumps(value)
```
Apoorvgarg-creator, line 377: Quick sanity question: when the archive already has a JSONB column serialized as a string (because some older export wrote it that way), this branch leaves it as a string and the `::jsonb` cast in `_build_insert` parses it on the way in. When it comes through as a Python dict or list (typical case), `json.dumps` round trips it. That covers both shapes, which is good.

The round trip does mean values like `NaN`, `Infinity`, or non string dict keys would now silently coerce or raise. If those shapes never appear in source data we're fine; if there's any chance they do (eval scorecards, model configs, anything user authored), it's cheaper to find out here than at row N of a 500k row import. Worth one focused test feeding representative `agents` / `scorecards` rows through `_coerce_value`.

observal_cli/cmd_migrate.py
```diff
@@ -334,9 +372,18 @@ def _coerce_value(value: object, pg_type: str) -> object:
...
+    # asyncpg requires JSON/JSONB values as serialized strings
+    if pg_type in ("json", "jsonb") and not isinstance(value, str):
+        return json.dumps(value)
     return value
 
 
+# Columns that are NOT NULL with JSON defaults — used when archive has NULL for these.
+_NOTNULL_JSON_DEFAULTS: dict[str, dict[str, str]] = {
+    "agent_versions": {"models_by_ide": "{}"},
+}
```
Apoorvgarg-creator, line 384: Two thoughts on this:

1. The defaults already exist in the schema (the column was declared `NOT NULL DEFAULT '{}'::jsonb` in the alembic migration that added it). Reading `column_default` from `information_schema.columns` once per table avoids this map drifting from the schema as new NOT NULL JSONB columns are added; today it's only `models_by_ide`, but the next addition will silently fall through this defaulting path.

2. The deeper question is why the export wrote `NULL` for a column that is `NOT NULL` on the source. Either the source has rows that predate the constraint (in which case there's data cleanup debt on the source that we're papering over here), or the export path is dropping the value somewhere. A one line warning emitted when this default kicks in would surface the divergence without blocking the migration, and gives us a thread to pull on later.

observal_cli/cmd_migrate.py
```diff
@@ -389,6 +442,10 @@ async def _flush_batch(
             row_id = row.get("id", "unknown")
             rprint(f"[yellow]  FK violation in {table}, row {row_id}: {e.constraint_name}[/yellow]")
             skipped += 1
+        except asyncpg.UniqueViolationError as e:
+            row_id = row.get("id", "unknown")
+            rprint(f"[yellow]  Unique conflict in {table}, row {row_id}: {e.constraint_name}[/yellow]")
+            skipped += 1
```
Apoorvgarg-creator, line 448: The `_build_insert` query at line 398 is already `ON CONFLICT ("id") DO NOTHING`, so the only way to land in this handler is a unique constraint on a non PK column: slugs, `(org_id, name)` tuples, etc.

Swallowing those silently during a cross environment import means two distinct source rows can collapse into one on the target without anyone noticing, since the warning line scrolls past during a long run. Two options worth weighing:

1. Promote unique conflicts on non PK columns to the `warnings` list returned in `ImportResult`, so they show up in the final summary rather than only in terminal scroll.
2. Or accept the current behavior but include the conflicting column values in the warning, so an operator scanning logs after the fact knows what got dropped.

For the PK case specifically, that's already a no op via `ON CONFLICT`, so this handler should never fire for primary keys. Confirming that assumption with a comment would help the next reader.

Apoorvgarg-creator (commented):
Thanks for pulling this one apart. The import was visibly fragile before and this gets it most of the way there. The big shapes work: version tables in the insert order, idempotent re-import via `ON CONFLICT`, JSON coercion at the asyncpg boundary, and graceful handling of older archives. The PR description also does a good job calling out the gaps you didn't fix (org remapping coverage, traces visibility), which is exactly the right framing for a tool that's going to run against real data.

A few themes in the inline comments worth pulling out:

1. `DEFERRED_VERSION_BACKFILL` plus the `deferred_tables` map look like a half landed earlier approach. The actual cycle break is `session_replication_role = 'replica'`. Worth picking one strategy and deleting the other, because the two have very different side effects on triggers.

2. `session_replication_role = 'replica'` is a bigger hammer than the comment claims. It disables every user trigger on the session, not just FK checks, and the reset on line 861 isn't in a `finally` so any mid-import exception leaves it set. If we can lean on deferrable constraints for the listings to versions cycle, that's the narrower fix and probably what the `DEFERRED_VERSION_BACKFILL` scaffolding was aiming at originally.

3. `org_columns = {'org_id', 'owner_org_id'}` and `_NOTNULL_JSON_DEFAULTS` are both hardcoded mirrors of the schema. They will silently drift the first time someone adds an org FK or a NOT NULL JSONB column. Reading both from `information_schema` keeps the migration honest as the schema evolves.

4. Test coverage for the new behavior is thin. The count assertions got updated but the actually risky pieces (org rewriting, JSON coercion of edge values, unique conflict handling, NOT NULL default substitution) aren't exercised. Even a couple of unit tests against a mocked asyncpg connection would catch regressions when the next person touches this file.

5. Strongly consider making `migrate import` refuse to run, or print a hard warning, when the source archive contains org ids that don't exist on the target and `--org-id` was not passed. The 'Known Gaps' section describes the failure mode well, and the natural place to enforce that is at the entry point rather than asking the operator to read the docs.

None of this blocks the fix in scope, which is unblocking `observal migrate import` against a fresh target. Reasonable to merge once the trigger reset and the dead constant are cleaned up, with the schema introspection and tests landing as a follow up.

observal_cli/cmd_migrate.py
```diff
@@ -725,6 +799,34 @@ async def _import_archive(db_url: str, archive_path: Path) -> ImportResult:
...
+                                continue
+                            row = json.loads(line)
+                            src_id = row.get("id")
+                            if src_id and src_id != normalize_org_id:
+                                org_rewrite_map[src_id] = normalize_org_id
+                if org_rewrite_map:
+                    rprint(f"[dim]  Normalizing {len(org_rewrite_map)} source org(s) to: {normalize_org_id}[/dim]")
+
+            # Columns that hold org references and should be rewritten
+            org_columns = {"org_id", "owner_org_id"}
```
Apoorvgarg-creator, line 825: Hardcoding the rewrite column names is what the PR description's 'Known Gaps' already calls out as fragile, and I'd push back a little on shipping the `--org-id` flag with this limitation baked in.

For `--org-id` to be trustworthy as a hands off remap, this set needs to be derived from the schema: any column whose FK references `organizations.id`, regardless of column name. `information_schema.referential_constraints` joined with `key_column_usage` will give exactly that list at runtime. As written, a future table that uses (say) `tenant_org_id` or `parent_org_id` will silently not be rewritten, the import will succeed, and the operator won't know until they hit a UI that doesn't show their data.

Given this is the headline 'use the flag' code path for users self hosting against a fresh org, worth doing it properly the first time rather than carrying it as known debt.

observal_cli/cmd_migrate.py
```diff
@@ -725,6 +799,34 @@ async def _import_archive(db_url: str, archive_path: Path) -> ImportResult:
...
+
+            # Temporarily disable FK checks for clean bulk import
+            await conn.execute("SET session_replication_role = 'replica'")
```
Apoorvgarg-creator, line 828: Worth flagging that this isn't only an FK toggle. `session_replication_role = 'replica'` disables every non replica user trigger on the session: `updated_at` triggers, audit triggers, anything that maintains derived columns or search vectors, and the constraint triggers used for `DEFERRABLE` constraints. Whatever invariants those triggers were enforcing now bypass for the whole import window, silently.

Two follow ups worth considering:

1. If the real cycle is just the listings to versions circular FK, those are typically deferrable. `SET CONSTRAINTS ALL DEFERRED` inside a transaction is the narrower hammer, keeps `updated_at` and friends firing, and matches what the `DEFERRED_VERSION_BACKFILL` constant above was reaching for. It does require wrapping the inserts in an explicit transaction, which today we don't have.
2. On managed Postgres (RDS without `rds_superuser`, Cloud SQL, etc.) setting `session_replication_role` requires elevated role membership and can fail with a not obvious error. Worth either documenting the role requirement in `docs/cli/migrate.md` or catching that error here and pointing the operator at it.

At minimum, the 'Temporarily disable FK checks' comment is misleading and should be updated to describe what the statement actually does.

observal_cli/cmd_migrate.py
```diff
@@ -735,13 +837,29 @@ async def _import_archive(db_url: str, archive_path: Path) -> ImportResult:
...
+                    jsonl_path,
+                    col_types,
+                    org_rewrite_map=org_rewrite_map,
+                    org_columns=org_columns,
+                )
                 rows_inserted[table] = ins
                 rows_skipped[table] = sk
 
+            # Re-enable FK checks
+            await conn.execute("SET session_replication_role = 'origin'")
```
Apoorvgarg-creator, line 861: If anything between 828 and here raises (an `_insert_table` error, an unexpected `asyncpg` exception, a keyboard interrupt), this reset never runs. The `finally` on line 890 only closes the connection.

In practice the connection is freshly acquired in this function and closed immediately after, so the leaked session state doesn't bleed across calls today. But that's a coincidence of the current structure, not a guarantee, and the moment someone reuses this with a pool or moves the connection acquisition up a level the bug becomes real. Wrap lines 828 to 861 in `try / finally`, or factor it into a small async context manager, so the reset is paired with the set unconditionally.

tests/test_migrate.py
```diff
@@ -160,8 +160,8 @@ def test_mixed_row(self):
 
 
 class TestConstants:
-    def test_insert_order_has_33_entries(self):
-        assert len(INSERT_ORDER) == 33
+    def test_insert_order_has_39_entries(self):
+        assert len(INSERT_ORDER) == 39
```
Apoorvgarg-creator, line 164: Bumping the count to 39 is correct, but it's worth noting that the rest of the test surface in this file didn't grow to match the actual behavior change in this PR. The risky pieces of the diff (org rewriting, JSON coercion of dict / list values, NOT NULL default substitution, UniqueViolationError handling) aren't exercised anywhere.

A few targeted unit tests against a `MagicMock` or `AsyncMock` of `asyncpg.Connection` would be cheap insurance:

- `_insert_table` with `org_rewrite_map` populated, asserting that `conn.execute` is called with the rewritten UUID.
- `_coerce_value` on `dict`, `list`, pre serialized `str`, and `None` JSON inputs.
- `_flush_batch` with a row that triggers `UniqueViolationError` from a stub, asserting `skipped` increments and the loop continues.

These would also serve as executable documentation for the next person who has to touch this code.

PR #817, opened 2026-05-10
test: assert real behaviour in graphql, worker, and listing detail tests

### LINKED ISSUE #806: Ongoing: fix hollow tests that pass without testing real logic
We have good test coverage by count but some tests pass without actually verifying the behaviour they claim to test. This is an ongoing issue to track and fix them.

Common patterns to look for:

- `assert x is not None` with no further assertions on the value
- `assert r.status_code == 200` with no check on the response body
- Tests that mock out the thing being tested (e.g. patching `resolve_listing` and then asserting the mock was called)
- GraphQL schema tests that only check the schema object exists (`assert schema is not None`) without running a query
- Tests that import a module and assert the import succeeded as the only assertion

## Known examples to fix

- `tests/test_graphql_phase6.py` - `test_schema_exists`, `test_has_query_type`, `test_has_subscription_type` only assert the schema object is not None, they never execute a query
- `tests/test_worker_phase5.py` - several tests only check that keys exist in the docker-compose YAML rather than testing actual worker behaviour
- `tests/test_listing_detail_access.py` - `test_sees_approved` checks status 200 but never validates the response body matches the listing

## How to contribute

1. Pick a test file
2. Find a test that passes trivially without testing real logic
3. Rewrite it to assert something meaningful about the actual output
4. Run `make test` to confirm it still passes

Please read our [CONTRIBUTING.md](../CONTRIBUTING.md) before submitting a PR.

You do not need to be assigned this issue to work on it.

dimasd-angga:
## Summary
Three tests named in #806 passed even when the underlying behaviour was broken: `test_schema_exists` only checked the schema object existed, `test_has_functions` only checked the list was non-empty, and the listing detail tests only asserted a 200. This PR rewrites them so they fail when resolvers, arq registrations, the worker compose command, or the listing response body actually regress. Scope is intentionally limited to the three files called out in the issue — the broader hollow-test sweep is ongoing.

## Changes
- `tests/test_graphql_phase6.py`: replace `test_schema_exists` / `test_has_query_type` / `test_has_subscription_type` with assertions on required `Query` field names (`traces`, `trace`, `span`, `mcpMetrics`, `overview`, `trends`), required `Subscription` fields (`traceCreated`, `spanCreated`, `sessionUpdated`), the return type of `Query.trace`, and an end-to-end `schema.execute` of a `traces` query that pins the response shape.
- `tests/test_worker_phase5.py`: replace `test_has_functions` with a check that `run_eval` is registered in `WorkerSettings.functions` by identity and by `__name__`. Tighten `test_calls_eval_and_publishes` to assert the published payload (`agent_id`, `trace_id`, `scores_written`) and the `run_eval_on_trace` call args. Add `test_iterates_traces_when_no_trace_id` for the fan-out path and harden `test_handles_eval_error` to assert nothing is published on failure.
- `tests/test_worker_phase5.py`: replace `test_worker_service_exists` with `test_worker_service_runs_arq_with_worker_settings` (asserts the compose command actually invokes `arq` against `WorkerSettings`) and add `test_worker_depends_on_redis_healthy` so a missing healthcheck dependency fails the suite.
- `tests/test_listing_detail_access.py`: in both `TestUnauthenticatedAccess.test_sees_approved` and `TestNonOwnerRegularUser.test_sees_approved`, validate the JSON body's `id` / `name` / `version` / `status` against the mocked listing instead of only checking `r.status_code == 200`.

## Testing
- `make test` — full suite passes locally with the rewritten assertions.

## Notes
- Other hollow tests across the suite are out of scope for this PR per the issue description ("ongoing"); this only covers the three files explicitly named.
- The new GraphQL execute test assumes the `traces` resolver delegates to `query_traces` in `api.graphql`; if that indirection changes, the patch target needs to move with it.

Closes #806

### DIFF
--- a/tests/test_graphql_phase6.py
+++ b/tests/test_graphql_phase6.py
@@ -164,14 +164,47 @@ class TestDataLoaders:
 
 
 class TestSchema:
-    def test_schema_exists(self):
-        assert schema is not None
+    def test_schema_exposes_required_query_fields(self):
+        # Renames or removals of these resolvers would break the dashboard,
+        # so the schema must keep advertising them.
+        query_type = schema._schema.query_type
+        assert query_type is not None
+        field_names = set(query_type.fields.keys())
+        for required in ("traces", "trace", "span", "mcpMetrics", "overview", "trends"):
+            assert required in field_names, f"Query.{required} missing from schema; got {sorted(field_names)}"
 
-    def test_has_query_type(self):
-        assert schema._schema.query_type is not None
+    def test_query_trace_returns_trace_type(self):
+        # Guards against the resolver being retyped (e.g. to a connection or scalar).
+        query_type = schema._schema.query_type
+        trace_field = query_type.fields["trace"]
+        # GraphQL nullable Trace serialises as type name "Trace"
+        assert "Trace" in str(trace_field.type)
 
-    def test_has_subscription_type(self):
-        assert schema._schema.subscription_type is not None
+    def test_schema_exposes_required_subscription_fields(self):
+        sub_type = schema._schema.subscription_type
+        assert sub_type is not None
+        field_names = set(sub_type.fields.keys())
+        for required in ("traceCreated", "spanCreated", "sessionUpdated"):
+            assert required in field_names, f"Subscription.{required} missing from schema; got {sorted(field_names)}"
+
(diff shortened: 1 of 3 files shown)


ShaanNarendran:
Please use the PR template

PR #886, opened 2026-05-12
Add guidance for supporting new IDEs in contributing.md

rithvik-p:


ShaanNarendran:
I'd like for this to be apart of your other pr as a separate commit since it doesn't make sense to split them like this and have a pr for such a small change.

PR #888, opened 2026-05-12
feat: insights migration

naraen-ram:
## Purpose / Description

Add the four insight-related PostgreSQL tables (`insight_reports`, `insight_session_facets`, `insight_session_meta`, `insight_meta_cache`) to the Observal CLI migration system. This enables shallow-copy export/import of insight data alongside the existing tables, ensuring FK ordering is respected and JSONB columns are properly cast during export.

Previously, running `observal migrate export` and `import` would silently skip all insight data since the tables weren't in `INSERT_ORDER`. Operators migrating between environments would lose all generated insight reports.

### 1. Constants Update (`observal_cli/cmd_migrate.py`)

**INSERT_ORDER** (39 → 43 entries):
- Added `insight_meta_cache`, `insight_session_facets`, `insight_session_meta`, `insight_reports` as Tier 12
- `insight_reports` is last among the four (most FK dependencies: agents, users, self-FK)
- Self-referential `previous_report_id` FK is handled by `session_replication_role = 'replica'` during import (same pattern as listing/version circular FKs)

**JSONB_COLUMNS** (4 new entries):
| Table | JSONB Columns |
|-------|--------------|
| `insight_reports` | `metrics`, `narrative`, `aggregated_data` |
| `insight_session_facets` | `facets` |
| `insight_session_meta` | `meta` |
| `insight_meta_cache` | `session_metas` |

### 2. Alembic Migration 0005 (`observal-server/alembic/versions/0005_add_insight_tables.py`)

- Creates `insight_reports`, `insight_session_facets`, `insight_session_meta` tables
- Idempotent: uses `has_table()` checks before each `create_table()` (handles race with `Base.metadata.create_all` on fresh installs)
- Creates indexes on `insight_reports.agent_id` and `insight_reports.triggered_by` (prevents full table scans on CASCADE deletes)
- Fixes `insight_meta_cache` FK to add missing `ondelete="CASCADE"`
- `downgrade()` drops tables in reverse creation order + cleans up enum type
- All timestamp columns use `server_default=sa.func.now()`

### 3. Model Fixes (`observal-server/models/`)

- Added `ondelete="CASCADE"` / `ondelete="SET NULL"` to insight model FKs (matching the Alembic DDL)
- Added `index=True` on `insight_reports.agent_id` and `triggered_by`

### 4. Test Updates (`tests/test_migrate.py`)

- Count assertion: 39 → 43
- Added 5 FK pairs to both `KNOWN_FK_PAIRS` and `ALL_FK_PAIRS`
- Existing Property 3 (JSONB cast) and Property 8 (FK ordering) automatically cover new tables

## How Has This Been Tested?

- Exported from dev.observal.io (43 tables, 665 rows including insight data)
- Imported into a fresh local instance with `--org-id` — 661 inserted, 4 skipped (expected demo conflicts)
- Verified insight data landed: 10 `insight_reports`, 6 `insight_session_facets`, 7 `insight_session_meta`
- Confirmed insight data is queryable via direct SQL after import
- Ruff check + format pass
- Note: Insights UI requires enterprise feature flag to display reports (data is present but API gates access)

## Execution Example

```
# Export from dev (includes insight tables)
observal migrate export --db-url <source-postgres-url> --output ./migration/dev-export.tar.gz

# Import into local with org remapping
observal migrate import --db-url <target-postgres-url> --archive ./migration/dev-export.tar.gz --org-id <target-org-id>
```

> [!NOTE]
> **Insights UI Visibility:** After import, the Insights page on localhost shows "No reports generated yet" despite the data being present in the database. This is because the `/api/v1/insights` endpoint is gated behind an enterprise feature flag — the API returns `"Insights is an enterprise feature"` when the flag is not enabled. The data is confirmed imported correctly via direct DB query:
>
> ```sql
> SELECT 'insight_reports' as tbl, count(*) FROM insight_reports
> UNION ALL SELECT 'insight_session_facets', count(*) FROM insight_session_facets
> UNION ALL SELECT 'insight_session_meta', count(*) FROM insight_session_meta;
> ```
>
> | tbl | count |
> |-----|-------|
> | insight_reports | 10 |
> | insight_session_facets | 6 |
> | insight_session_meta | 7 |
>
> This confirms the migration works correctly — the UI limitation is an access control gate, not a data issue.

## Checklist

- [ ] UI changes: include screenshots of all affected screens (N/A - backend only)

### DIFF
--- a/observal-server/alembic/versions/0004_add_insight_tables.py
+++ b/observal-server/alembic/versions/0004_add_insight_tables.py
@@ -1,0 +1,140 @@
+"""Add insight_reports, insight_session_facets, insight_session_meta tables.
+
+Revision ID: 0004
+Revises: 0003
+Create Date: 2026-05-10
+"""
+
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+from alembic import op
+
+revision = "0004"
+down_revision = "0003"
+branch_labels = None
+depends_on = None
+
+# Enum type name — managed explicitly to ensure clean downgrade
+_STATUS_ENUM_NAME = "insight_report_status"
+
+
+def upgrade() -> None:
+    bind = op.get_bind()
+    inspector = sa.inspect(bind)
+
+    # --- Fix insight_meta_cache FK (0003 omitted ondelete CASCADE) ---
+    if inspector.has_table("insight_meta_cache"):
+        # Drop and recreate the FK constraint with CASCADE
+        fks = inspector.get_foreign_keys("insight_meta_cache")
+        for fk in fks:
+            if fk.get("referred_table") == "agents" and "agent_id" in fk.get("constrained_columns", []):
+                op.drop_constraint(fk["name"], "insight_meta_cache", type_="foreignkey")
+                op.create_foreign_key(
+                    fk["name"],
+                    "insight_meta_cache",
+                    "agents",
+                    ["agent_id"],
+                    ["id"],
+                    ondelete="CASCADE",
+                )
+                break
+
+    # --- insight_reports ---
+    if not inspector.has_table("insight_reports"):
+        # Create enum type explicitly (checkfirst handles re-runs)
+        status_enum = sa.Enum("pending", "running", "completed", "failed", name=_STATUS_ENUM_NAME)
+        status_enum.create(bind, checkfirst=True)
+
+        op.create_table(
+            "insight_reports",
+            sa.Column(
+                "id", postgresql.UUID(as_uuid=True), primary_key=True, server_default=sa.text("gen_random_uuid()")
+            ),
+            sa.Column(
+                "agent_id",
+                postgresql.UUID(as_uuid=True),
+                sa.ForeignKey("agents.id", ondelete="CASCADE"),
+                nullable=False,
+            ),
+            sa.Column(
+                "triggered_by",
+                postgresql.UUID(as_uuid=True),
+                sa.ForeignKey("users.id", ondelete="SET NULL"),
+                nullable=True,
+            ),
+            sa.Column(
+                "status",
+                sa.Enum("pending", "running", "completed", "failed", name=_STATUS_ENUM_NAME, create_type=False),
+                nullable=False,
+                server_default="pending",
+            ),
+            sa.Column("period_start", sa.DateTime(timezone=True), nullable=False),
+            sa.Column("period_end", sa.DateTime(timezone=True), nullable=False),
+            sa.Column("metrics", postgresql.JSON(), nullable=True),
+            sa.Column("narrative", postgresql.JSON(), nullable=True),
+            sa.Column("sessions_analyzed", sa.Integer(), nullable=False, server_default="0"),
+            sa.Column("llm_model_used", sa.String(255), nullable=True),
+            sa.Column("error_message", sa.Text(), nullable=True),
+            sa.Column("started_at", sa.DateTime(timezone=True), server_default=sa.func.now(), nullable=False),
+            sa.Column("completed_at", sa.DateTime(timezone=True), nullable=True),
+            sa.Column("created_at", sa.DateTime(timezone=True), server_default=sa.func.now(), nullable=False),
+            sa.Column(
+                "previous_report_id",
+                postgresql.UUID(as_uuid=True),
+                sa.ForeignKey("insight_reports.id", ondelete="SET NULL"),
+                nullable=True,
+            ),
+            sa.Column("aggregated_data", postgresql.JSON(), nullable=True),
+            sa.Column("report_version", sa.Integer(), nullable=False, server_default="1"),
+        )
+        op.create_index("ix_insight_reports_agent_id", "insight_reports", ["agent_id"])
+        op.create_index("ix_insight_reports_triggered_by", "insight_reports", ["triggered_by"])
+
+    # --- insight_session_facets ---
+    if not inspector.has_table("insight_session_facets"):
+        op.create_table(
+            "insight_session_facets",
+            sa.Column(
+                "id", postgresql.UUID(as_uuid=True), primary_key=True, server_default=sa.text("gen_random_uuid()")
+            ),
+            sa.Column(
+                "agent_id",
+                postgresql.UUID(as_uuid=True),
+                sa.ForeignKey("agents.id", ondelete="CASCADE"),
+                nullable=False,
+            ),
+            sa.Column("session_id", sa.Text(), nullable=False),
+            sa.Column("extracted_at", sa.DateTime(timezone=True), server_default=sa.func.now(), nullable=False),
+            sa.Column("model_used", sa.String(255), nullable=True),
+            sa.Column("facets", postgresql.JSON(), nullable=False),
+            sa.UniqueConstraint("agent_id", "session_id", name="uq_session_facets_agent_session"),
+        )
+
+    # --- insight_session_meta ---
+    if not inspector.has_table("insight_session_meta"):
+        op.create_table(
+            "insight_session_meta",
+            sa.Column(
+                "id", postgresql.UUID(as_uuid=True), primary_key=True, server_default=sa.text("gen_random_uuid()")
+            ),
+            sa.Column(
+                "agent_id",
+                postgresql.UUID(as_uuid=True),
+                sa.ForeignKey("agents.id", ondelete="CASCADE"),
+                nullable=False,
+            ),
+            sa.Column("session_id", sa.Text(), nullable=False),
+            sa.Column("computed_at", sa.DateTime(timezone=True), server_default=sa.func.now(), nullable=False),
+            sa.Column("meta", postgresql.JSON(), nullable=False),
+            sa.UniqueConstraint("agent_id", "session_id", name="uq_session_meta_agent_session"),
+        )
+
+
+def downgrade() -> None:
+    op.drop_table("insight_session_meta")
+    op.drop_table("insight_session_facets")
+    op.drop_table("insight_reports")
+
+    # Drop the enum type created by upgrade
+    sa.Enum(name=_STATUS_ENUM_NAME).drop(op.get_bind(), checkfirst=True)
--- a/observal-server/alembic/versions/0005_add_insight_tables.py
+++ b/observal-server/alembic/versions/0005_add_insight_tables.py
@@ -1,3 +1,6 @@
+# SPDX-FileCopyrightText: 2026-present Observal (BlazeUp AI LLP)
+# SPDX-License-Identifier: AGPL-3.0-only
+
 """Add insight_reports, insight_session_facets, insight_session_meta tables.
 
 Revision ID: 0005
--- a/observal-server/alembic/versions/0005_add_insight_tables.py
+++ b/observal-server/alembic/versions/0005_add_insight_tables.py
@@ -1,8 +1,8 @@
 """Add insight_reports, insight_session_facets, insight_session_meta tables.
 
-Revision ID: 0004
-Revises: 0003
-Create Date: 2026-05-10
+Revision ID: 0005
+Revises: 0004
+Create Date: 2026-05-12
 """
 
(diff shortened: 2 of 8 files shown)


naraen-ram:
**CI Note:** The `security-scan` failure is pre-existing and unrelated to this PR. Bandit excludes `alembic/` per `pyproject.toml` config, and running bandit on our changed files (`cmd_migrate.py`, models) produces zero findings.

observal-server/alembic/versions/0004_add_insight_tables.py
```diff
@@ -0,0 +1,140 @@
+"""Add insight_reports, insight_session_facets, insight_session_meta tables.
+
+Revision ID: 0004
```
Haz3-jolt: 0004 already exists, you have to regenerate alembic

Haz3-jolt (changes requested):

observal-server/alembic/versions/0005_add_insight_tables.py
```diff
@@ -0,0 +1,143 @@
...
+
+from alembic import op
+
+revision = "0005"
+down_revision = "0004"
+branch_labels = None
+depends_on = None
+
+# Enum type name — managed explicitly to ensure clean downgrade
+_STATUS_ENUM_NAME = "insight_report_status"
```
Apoorvgarg-creator, line 22: I hope if you are creating Enum at the top of the file, it is just used inside the file only and not shared between different files.

observal_cli/cmd_migrate.py
```diff
@@ -284,10 +299,34 @@ def _sha256_file(path: Path) -> str:
...
+    """
+    import sys
+
+    if sys.version_info >= (3, 12):
+        tar.extractall(dest, filter="data")
+    else:
+        # Manual path traversal protection for Python < 3.12
+        for member in tar.getmembers():
+            member_path = (dest / member.name).resolve()
+            if not str(member_path).startswith(str(dest.resolve())):
```
Apoorvgarg-creator: I think this `startswith` check has the classic prefix-string hole. If `dest` resolves to `/tmp/staging` and a tar member's name resolves to `/tmp/staging-malicious/foo` (via `../../staging-malicious/foo`), the string `/tmp/staging-malicious/foo` does start with `/tmp/staging`, so it slips through.

`Path.is_relative_to()` is available on 3.9+ (and we already gate this whole branch on `< 3.12`), so this is cleaner and correct:

```python
member_path = (dest / member.name).resolve()
if not member_path.is_relative_to(dest.resolve()):
    raise ValueError(...)
```

Low severity since `migrate import` is admin-gated, but the function's whole job is path-traversal protection, so worth tightening.

observal_cli/cmd_migrate.py
```diff
@@ -262,17 +271,23 @@ def _require_admin() -> None:
 
 
 def _build_select(table: str, columns: list[str]) -> str:
-    """Build SELECT query, casting JSONB columns to ::text."""
+    """Build SELECT query, casting JSONB columns to ::text.
+
+    Table names are validated against INSERT_ORDER to prevent SQL injection.
```
Apoorvgarg-creator: Small thing on the docstring wording: the existing callers pass `table` from `INSERT_ORDER` (loop variable) and `columns` from `stmt.get_attributes()` on a prepared statement, neither of which is user-controlled, so this isn't really an injection fix, it's a defense-in-depth assertion. Worth rephrasing to something like "Guards against accidental misuse by callers passing unknown tables", otherwise a future reader will go hunting for the user input that doesn't exist.

observal_cli/cmd_migrate.py
```diff
@@ -957,11 +1004,11 @@ async def _import_archive(db_url: str, archive_path: Path, normalize_org_id: str
                 if "owner_org_id" not in tbl_cols:
                     continue
                 result = await conn.execute(
-                    f"UPDATE {tbl} SET owner_org_id = u.org_id "
+                    f'UPDATE "{tbl}" SET "owner_org_id" = u."org_id" '
                     f"FROM users u "
```
Apoorvgarg-creator: Quoting got partly applied here: `"{tbl}"`, `"owner_org_id"`, `u."org_id"`, `u."id"` are all quoted but `FROM users u` (and the `u.` alias prefix) is not. Probably worth either quoting `"users"` here too or dropping the quotes elsewhere in this statement so the style is consistent. Mixed quoting is harmless to Postgres but invites "why is this one different?" comments in future reviews.

observal-server/models/insight_meta_cache.py
```diff
@@ -23,7 +23,9 @@ class InsightMetaCache(Base):
     __table_args__ = (UniqueConstraint("agent_id", "period_start", "period_end", name="uq_meta_cache_agent_period"),)
 
     id: Mapped[uuid.UUID] = mapped_column(UUID(as_uuid=True), primary_key=True, default=uuid.uuid4)
-    agent_id: Mapped[uuid.UUID] = mapped_column(UUID(as_uuid=True), ForeignKey("agents.id"), nullable=False)
+    agent_id: Mapped[uuid.UUID] = mapped_column(
+        UUID(as_uuid=True), ForeignKey("agents.id", ondelete="CASCADE"), nullable=False
```
Apoorvgarg-creator, line 27: One thing not covered by either the migration or this model change: if someone is running a server that does its DDL through `Base.metadata.create_all` (the lifespan path in `observal-server/main.py:89`) and already has `insight_meta_cache` from a previous boot, the new `ondelete="CASCADE"` here won't be applied. `create_all` only creates missing objects, it doesn't alter existing FKs.

Users on alembic get the fix via 0005. Users on `create_all` need either to drop+recreate the constraint at startup (similar to the `_ensure_columns` block in `main.py`) or to manually apply 0005. Might be worth a one-line note in the PR description or a follow-up issue.

observal-server/alembic/versions/0005_add_insight_tables.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+def upgrade() -> None:
+    bind = op.get_bind()
+    inspector = sa.inspect(bind)
+
```
Apoorvgarg-creator, line 28: Building on the existing comment about the enum: the migration creates the Postgres type as `insight_report_status` (snake_case), but `observal-server/models/insight_report.py:35` declares the column as `Enum(InsightReportStatus)` without `name=`, so SQLAlchemy will derive the type name from the class `__name__` lowercased, which becomes `insightreportstatus` (no underscores). On a fresh install, the `Base.metadata.create_all` path will create a different type from what 0005 creates.

The cleanest fix is to add `name="insight_report_status"` to the model's `Enum(...)` call so both paths produce the same Postgres type name. Worth doing in this PR rather than as a follow-up, since renaming a Postgres enum later is its own migration.

observal-server/alembic/versions/0005_add_insight_tables.py
```diff
@@ -0,0 +1,143 @@
...
+                    ["agent_id"],
+                    ["id"],
+                    ondelete="CASCADE",
+                )
+                break
+
+    # --- insight_reports ---
+    if not inspector.has_table("insight_reports"):
+        # Create enum type explicitly (checkfirst handles re-runs)
+        status_enum = sa.Enum("pending", "running", "completed", "failed", name=_STATUS_ENUM_NAME)
```
Apoorvgarg-creator, line 51: The FK lookup-and-recreate is fine, but I'd add a small guard: if the existing FK already has `ondelete="CASCADE"` (e.g. because someone manually fixed it, or because the table was created via `Base.metadata.create_all` after the model change lands), this block will still drop and recreate it. Functionally a no-op, but it's a brief moment where the constraint doesn't exist. Checking `fk.get("options", {}).get("ondelete")` before recreating would skip that no-op.

Apoorvgarg-creator (commented):
Looks solid overall, just a few things to look at before merge. Most of the inline comments are on the export/import hardening in `cmd_migrate.py` rather than the migration itself.

observal-server/alembic/versions/0005_add_insight_tables.py
```diff
@@ -0,0 +1,143 @@
...
+            sa.Column(
+                "previous_report_id",
+                postgresql.UUID(as_uuid=True),
+                sa.ForeignKey("insight_reports.id", ondelete="SET NULL"),
+                nullable=True,
+            ),
+            sa.Column("aggregated_data", postgresql.JSON(), nullable=True),
+            sa.Column("report_version", sa.Integer(), nullable=False, server_default="1"),
+        )
+        op.create_index("ix_insight_reports_agent_id", "insight_reports", ["agent_id"])
```
Apoorvgarg-creator, line 98: The PR description mentions that `previous_report_id`'s self-FK is handled by `session_replication_role = 'replica'` during import, which is true for the import path, but worth a brief docstring note here so the next person editing this migration doesn't see the self-FK and wonder if INSERT ordering needs to handle it. Just a comment, no code change.

observal-server/alembic/versions/0005_add_insight_tables.py
```diff
@@ -0,0 +1,143 @@
+# SPDX-FileCopyrightText: 2026-present Observal (BlazeUp AI LLP)
+# SPDX-License-Identifier: AGPL-3.0-only
+
+"""Add insight_reports, insight_session_facets, insight_session_meta tables.
+
+Revision ID: 0005
+Revises: 0004
+Create Date: 2026-05-12
```
Apoorvgarg-creator, line 8: Minor: the PR description says "Alembic Migration 0004" multiple times but the file is `0005_add_insight_tables.py` with `revision = "0005"`. `0004` is the avatar-url migration already on main. Just a description update so reviewers aren't confused looking for a 0004 file in this PR.

tests/test_migrate.py
```diff
@@ -164,8 +164,8 @@ def test_mixed_row(self):
 
 
 class TestConstants:
-    def test_insert_order_has_39_entries(self):
-        assert len(INSERT_ORDER) == 39
+    def test_insert_order_has_43_entries(self):
```
Apoorvgarg-creator, line 167: Tests cover the constants and the FK-pair properties, but there's no test that actually exercises the new migration 0005 against a database (upgrade then downgrade). If the alembic suite elsewhere already does a full upgrade-from-zero loop that'd catch breakage; if not, a single round-trip test would catch the self-FK and the enum-drop ordering, which are the two failure modes most likely to bite later.

observal-server/alembic/versions/0005_add_insight_tables.py
naraen-ram: Yes, it's a file local private constant. It isn't imported anywhere else.

naraen-ram:
## Execution Example

```bash
# Getting the target org id
docker exec docker-observal-db-1 psql -U postgres -d observal -c "SELECT id FROM organizations LIMIT 1;" -t

# Phase 1: Shallow copy (PostgreSQL)
observal migrate export \
  --db-url <source-postgres-url> \
  --output ./migration/dev-export.tar.gz

observal migrate validate \
  --archive ./migration/dev-export.tar.gz

# Import
observal migrate import \
  --db-url <target-postgres-url> \
  --archive ./migration/dev-export.tar.gz \
  --org-id <target-org-id>

# Phase 2: Deep copy (ClickHouse telemetry)
observal migrate export-telemetry \
  --clickhouse-url <source-clickhouse-url> \
  --manifest ./migration/dev-export.manifest.json \
  --output-dir ./migration/telemetry

# Import
observal migrate import-telemetry \
  --clickhouse-url <target-clickhouse-url> \
  --input-dir ./migration/telemetry \
  --project-id <target-org-id>

# Validate
observal migrate validate-telemetry \
  --input-dir ./migration/telemetry \
  --clickhouse-url <target-clickhouse-url>
```

observal-server/alembic/versions/0005_add_insight_tables.py
ShaanNarendran: Alembic may need to be changed, since main might be at 0007 by the time this pr is ready for merge

PR #921, opened 2026-05-13
fix(deploy): fix curl|bash interactive prompts

ShaanNarendran:
## Purpose / Description

When running `curl -fsSL ... | bash`, stdin is the pipe from curl — so all `read` prompts in the install/setup scripts immediately get EOF and exit. This is why the installer aborts at the first prompt.

## Approach

Redirect interactive reads from `/dev/tty` so terminal input works regardless of how the script was invoked. This is the standard pattern used by Docker, rustup, nvm, etc.

## How Has This Been Tested?

- Reproduced the failure: `curl ... | bash` exits at first prompt
- After fix: `read </dev/tty` reads from the terminal correctly

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,15 +122,32 @@ See [CHANGELOG.md](CHANGELOG.md) for recent updates.
 
 ## Quick start
 
-See [SETUP.md](SETUP.md) for the full setup guide.
+### One-line install (recommended)
+
+```bash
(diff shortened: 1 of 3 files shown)


Haz3-jolt:
move LICENSE to bottom of readme as well

PR #925, opened 2026-05-13
fix(security): validate session IDs before SQL interpolation (OBSV-SEC-029)

Haz3-jolt:
## Summary

Fixes SQL injection in `ee/observal_insights/shim_enrichment.py` where user-controlled session IDs were interpolated directly into a ClickHouse query string.

## The problem

```python
# BEFORE — vulnerable
ids_str = ", ".join(f"’{sid}’" for sid in session_ids)
sql = f"WHERE metadata[’session.id’] IN ({ids_str})"
```

A session ID like `’) OR 1=1 --` breaks out of the string literal. An attacker with control over telemetry session IDs could manipulate ClickHouse queries in the shim enrichment pipeline.

## The fix

```python
# AFTER — safe
_SESSION_ID_SAFE_RE = re.compile(r"^[a-zA-Z0-9_\-\.]{1,256}$")

def _sanitize_session_ids(session_ids: list[str]) -> list[str]:
    """Return only session IDs matching the safe format. Drops unsafe ones."""
    ...

session_ids = _sanitize_session_ids(session_ids)
if not session_ids:
    return {}
```

IDs that fail validation are logged and dropped. If all IDs are invalid the ClickHouse query is skipped entirely.

## Files changed

- `ee/observal_insights/shim_enrichment.py` — validation + sanitization
- `tests/test_shim_phase3.py` — 7 new tests covering injection attempts, mixed lists, boundary lengths

## Testing

```
51 passed in 0.30s
```

### DIFF
--- a/ee/observal_insights/shim_enrichment.py
+++ b/ee/observal_insights/shim_enrichment.py
@@ -15,6 +15,7 @@ Other IDEs will have no spans, and all functions handle that gracefully.
 
 from __future__ import annotations
 
+import re as _re
 from datetime import UTC
 
 import structlog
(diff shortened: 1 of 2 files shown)


ShaanNarendran (approved):

ShaanNarendran:
LGTM, if it's been physically tested once then we can merge

PR #981, opened 2026-05-14
fix(security): enforce password strength on account creation and change (OBSV-SEC-006)

Haz3-jolt:
## Purpose / Description

Weak passwords were accepted on `/auth/init` (initial admin creation) and `/auth/profile/password` (password change) with no minimum length, complexity, or common-password check.

## Fixes

* Fixes OBSV-SEC-006, password policy and account lockout controls are incomplete

## Approach

Added `_validate_password_strength(password)` to `api/routes/auth.py` enforcing:

- Minimum 12 characters
- At least one uppercase letter
- At least one digit
- At least one special character
- Not in a blocklist of 10 common weak passwords

Called in `init_admin` (when a password is provided) and `change_password`. Raises HTTP 422 with a descriptive message on failure. Existing users with weak passwords are unaffected until they attempt a password change.

## How Has This Been Tested?

`tests/test_sec006_password_policy.py`, 8 tests covering every policy rule and both HTTP endpoints. Full suite: 2021 passed, 24 skipped.

## Checklist

- [ ] UI changes: include screenshots of all affected screens (N/A, backend only)

### DIFF
--- a/observal-server/api/routes/auth.py
+++ b/observal-server/api/routes/auth.py
@@ -57,6 +57,45 @@ logger = logging.getLogger(__name__)
 
 router = APIRouter(prefix="/api/v1/auth", tags=["auth"])
 
+_COMMON_WEAK_PASSWORDS = frozenset(
+    {
+        # Passwords that fail the complexity rules anyway (kept for clarity)
+        "password",
+        "password1",
+        "password123",
+        "12345678",
+        "123456789",
+        "qwerty123",
+        "iloveyou",
+        "letmein123",
+        "welcome1",
+        "monkey123",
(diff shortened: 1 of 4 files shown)


observal-server/api/routes/auth.py
```diff
@@ -57,6 +57,35 @@
 
 router = APIRouter(prefix="/api/v1/auth", tags=["auth"])
 
+_COMMON_WEAK_PASSWORDS = frozenset(
```
ShaanNarendran: Is this an easter egg lol? Since, I think the other checks are enough, most of these fail the first check of 12 chars.

PR #984, opened 2026-05-14
fix(security): block password lifecycle routes in SSO-only mode (OBSV-SEC-028)

Haz3-jolt:
## Purpose / Description

`/auth/init` and `/auth/profile/password` did not check `SSO_ONLY`. In an SSO-only deployment these endpoints allowed local password creation and changes that should be disabled, bypassing the SSO requirement.

## Fixes

* Fixes OBSV-SEC-028, SSO-only and SCIM deactivation controls are inconsistent

## Approach

`require_password_auth` is already defined in `api/deps.py` and used on `/login` and `/token`. It raises HTTP 403 when `settings.SSO_ONLY=True`. Apply the same dependency to:

- `POST /api/v1/auth/init` (initial admin creation)
- `PUT /api/v1/auth/profile/password` (password change)

One-line change per route, identical pattern to the existing `/login` guard.

## How Has This Been Tested?

`tests/test_sec028_sso_guards.py`, 3 tests. Full suite: ~1900 passed.

## Checklist

- [ ] UI changes: include screenshots of all affected screens (N/A, backend only)

### DIFF
--- a/tests/test_sec028_sso_guards.py
+++ b/tests/test_sec028_sso_guards.py
@@ -1,0 +1,108 @@
+# SPDX-FileCopyrightText: 2026 Hari Srinivasan <harisrini21@gmail.com>
+# SPDX-License-Identifier: AGPL-3.0-only
+
+"""SEC-028: Password lifecycle routes must be blocked when SSO_ONLY=True."""
+
+from unittest.mock import AsyncMock, MagicMock, patch
+
+import pytest
+
+
+def _make_async_client():
+    from httpx import ASGITransport, AsyncClient
+
+    from api.ratelimit import limiter
+    from main import app
+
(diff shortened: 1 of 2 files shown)


tests/test_sec028_sso_guards.py
```diff
@@ -0,0 +1,109 @@
...
+from unittest.mock import AsyncMock, MagicMock, patch
+
+import pytest
+
+
+def _make_async_client():
+    from httpx import ASGITransport, AsyncClient
+
+    from api.ratelimit import limiter
+    from main import app
```
ShaanNarendran: nit: move common imports to the top since we repeat them, not sure if this affects functionality but afaik it shouldn't.

PR #991, opened 2026-05-14
Add the global command palette shortcut

adminlip:
Adds the existing command menu to the top-level provider so Cmd+K and Ctrl+K work across the app, and expands the palette with direct entries for traces, dashboard, settings, agents, and the latest traces from the telemetry query. The shortcut keeps the existing preventDefault behavior for the command chord, Esc closing is handled by the dialog, and navigation continues through the Next router.


Haz3-jolt:
Please use the PR template!

PR #1052, opened 2026-05-18
fix(web): add license field to package.json

Kaushik-Kumar-CEG:
Add `"license": "AGPL-3.0-only"` to `web/package.json`.

The field was missing — while not breaking (the package is `private: true`), compliance tools like `license-checker`, `reuse lint`, and SBOM generators use this field to identify the license from the manifest. Adding it keeps the frontend package consistent with `pyproject.toml` and `observal-server/pyproject.toml` which both declare `AGPL-3.0-only`.

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -2,6 +2,7 @@
   "name": "web",
   "version": "0.1.0",
   "private": true,
+  "license": "AGPL-3.0-only",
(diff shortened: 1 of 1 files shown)


ShaanNarendran:
@Kaushik-Kumar-CEG can this be merged?

PR #1063, opened 2026-05-19
fix(ee): fix SAML SSO

harishankar0301:
## Purpose / Description
SAML SSO: Okta & Azure AD (Entra ID) Compatibility

## Fixes

**Content-Type middleware exemption** (`observal-server/api/middleware/content_type.py`)
- Exempt `/api/v1/sso/saml` prefix from strict JSON content-type enforcement
- Required because the SAML ACS endpoint receives `application/x-www-form-urlencoded` POST from the IdP

**SAML settings corrections** (`ee/observal_server/services/saml.py`)
- Fix NameIDFormat to use the correct `SAML:1.1` URN for email format (the `2.0` variant doesn't exist in the spec)
- Disable `requestedAuthnContext` to avoid IdP rejection when the SP demands an AuthnContext the IdP doesn't support

**Request preparation fix** (`ee/observal_server/routes/sso_saml.py`)
- Derive scheme/host/port from `FRONTEND_URL` instead of the raw request object
- Fixes ACS URL mismatch when running behind a reverse proxy (nginx) where the container sees `localhost:8001` but the IdP expects the public URL
## Approach
Both Okta and Azure AD were rejecting SAML assertions due to:
1. ACS URL mismatch (container port vs public URL)
2. Invalid NameIDFormat URN
3. Unsupported AuthnContext requirement

## How Has This Been Tested?

- Tested against Okta developer tenant
- Tested against Azure AD (Entra ID) enterprise app
- Existing SAML tests pass (`make test`)

### DIFF
--- a/ee/README.md
+++ b/ee/README.md
@@ -12,7 +12,7 @@ Source-available enterprise features for Observal. This module is loaded only wh
 | Feature | Status | Description |
 |---------|--------|-------------|
 | Audit Logging | Implemented | Writes all admin and auth events to a ClickHouse `audit_log` table. Queryable and exportable as CSV. |
-| SAML 2.0 SSO | Stub (501) | Routes defined for SAML login, ACS callback, and SP metadata. Not yet implemented. |
+| SAML 2.0 SSO |Implemented | Working SAML SSO tested with Entra ID and Okta as IDPs|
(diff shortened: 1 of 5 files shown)


Haz3-jolt (approved):
LGTM

PR #1064, opened 2026-05-19
fix(security): redact nested structured secrets

tsitu0:
## Purpose / Description
This PR fixes a secrets redaction gap in structured telemetry payloads.

Before this change, `redact_dict(data, fields={...})` only reliably redacted direct string values under selected fields. That was not enough for realistic telemetry, because fields like `tool_input`, `tool_response`, `input`, or `output` can contain nested JSON structures: lists of messages, request objects, headers, params, metadata, or tuples.

That means a field could be selected for redaction, but secrets inside nested values could still be stored or returned through diagnostics.

This fix closes that gap at the redaction helper level, where all callers benefit from safer behavior without needing to remember to manually walk nested structures.

## Fixes
N/A

## Approach
The change updates `redact_dict` to treat a selected field as a whole structured payload.

- If it is a string, it redacts the string.
- If it is a dictionary, it recursively redacts string values inside it.
- If it is a list, it recursively redacts each item.
- If it is a tuple, it recursively redacts each item.

The change also keeps the previous targeted behavior for unrelated fields: unselected top-level fields are preserved, except that nested dictionaries/lists are still searched for selected field names such as `tool_input`.

This is the right tradeoff because it avoids over-redacting unrelated fields while still protecting secrets in fields that the system already considers sensitive.

The tests cover the important cases:
- nested lists inside selected fields
- nested dictionaries inside selected fields
- tuple values inside selected fields
- selected fields nested inside lists
- preservation of unrelated unselected values
- original input data is not mutated

I also split a few fake token literals in the test file so the repository’s secret scanner passes while preserving the same runtime test behavior.

## How Has This Been Tested?

Ran the focused redactor test suite:

```bash
uv run --with pytest pytest tests/test_secrets_redactor.py -q
```
Result:
```text
52 passed
```

Commit and push hooks also passed, including:
 - ruff
 - ruff format
 - SPDX checks
 - private key detection
 - secret scanning

The key issue was that telemetry payloads are often structured data, not plain strings. A tool call can send JSON like:
```json
{
  "tool_input": {
    "headers": {
      "Authorization": "Bearer ..."
    },
    "messages": [
      {
        "content": "password=..."
      }
    ]
  }
}
```

If redaction only handles the first level, sensitive values inside that structure can leak into telemetry storage. Recursive redaction for selected sensitive fields is a safer default.

Reference: OpenTelemetry recommends redacting or filtering sensitive data before telemetry is stored or exported. This PR applies that guidance to structured payloads: selected fields like `tool_input` can contain nested dictionaries, lists, headers, messages, or params, so the redactor needs to walk those containers instead of only checking direct string values.

https://opentelemetry.io/docs/security/handling-sensitive-data/

## Checklist

- [x] UI changes: N/A, no ui changes

### DIFF
--- a/observal-server/services/secrets_redactor.py
+++ b/observal-server/services/secrets_redactor.py
@@ -1,6 +1,7 @@
 # SPDX-FileCopyrightText: 2026 Hari Srinivasan <harisrini21@gmail.com>
 # SPDX-FileCopyrightText: 2026 Lokesh Selvam <lokeshselvam7025@gmail.com>
 # SPDX-FileCopyrightText: 2026 Vishnu Muthiah <vishnu.muthiah04@gmail.com>
+# SPDX-FileCopyrightText: 2026 tsitu0 <tomsitu0102@gmail.com>
 # SPDX-License-Identifier: AGPL-3.0-only
 
 """Secrets redactor for trace ingestion.
@@ -15,6 +16,7 @@ BEFORE storage in ClickHouse.  Designed to avoid over-stripping:
 """
 
 import re
+from typing import Any
 
 # ---------------------------------------------------------------------------
 # Known API key prefixes — near-zero false-positive rate.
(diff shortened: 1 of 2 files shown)


Haz3-jolt:
Thanks a lot! This was an excellent contribution!

PR #1152, opened 2026-05-20
fix(web): prevent auth guard flicker on direct URL access

### LINKED ISSUE #1088: fix(web): auth guard flicker shows protected page before redirect on direct URL access
## Problem

When navigating directly to a protected route (e.g. `http://localhost:3000/dashboard`) without being logged in, the page content flashes briefly before the auth guard redirects to `/login`. The effect is visible for roughly one render frame.

## Root cause

The auth guard (`src/components/layouts/auth-guard.tsx`) checks for an API key in localStorage, but on first render the check has not run yet. The component renders the protected children with no content check, then on the next render detects the missing key and redirects. This produces a one-frame flash of the protected UI.

## Expected behaviour

Protected pages should never render any content before auth state is confirmed. The user should see either a blank screen or a loading spinner immediately, with the redirect happening before any page content is painted.

## Fix

The auth guard should default to a loading/blank state until the localStorage check completes, not to rendering children. Concretely:

- On mount, start in a `checking` state that renders nothing (or a minimal spinner)
- After the first effect runs and confirms auth state, either allow render or redirect
- No protected content should ever be painted to the DOM before the check resolves

## Affected area

`web/src/components/layouts/auth-guard.tsx` and any equivalent role guard wrappers (`admin-guard.tsx`, `role-guard.tsx`).

d-kavinraja:
## Purpose / Description

When navigating directly to a protected route (e.g., `http://localhost:3000/dashboard`) without being logged in, the page content flashes briefly before the auth guard redirects to `/login`. The effect is visible for roughly one render frame.

**Root Cause:** The auth guard components (`AuthGuard` and `RoleGuard`) render children immediately without checking if auth/role state is confirmed. The authentication check happens in `useEffect`, causing a one-frame flicker of protected UI before the redirect takes effect.

## Fixes

* Fixes #1088

## Approach

Modified auth guards to wait for `ready` state confirmation before rendering any content. This ensures no protected content is ever painted to the DOM before auth is verified.

**Changes:**
- `AuthGuard` now returns `null` until auth state is confirmed
- `RoleGuard` now returns `null` until role is confirmed
- Prevents protected content from flashing before redirect

The guards previously rendered children immediately, causing a one-frame flash of protected UI before the useEffect redirect took effect. Now they wait for the ready state from their respective hooks before rendering any content.

### Test Environment
- **Browser:** Chrome/Edge on Windows
- **URL:** `http://localhost:3000`
- **Next.js Version:** 16.2.6
- **React Version:** 19.2.6

### Test Steps
1. Started frontend dev server: `cd web && pnpm dev`
2. Opened `http://localhost:3000`
3. Cleared auth token from Session Storage (DevTools → Application → Session Storage)
4. Navigated directly to protected route: `http://localhost:3000/dashboard`

### Expected Behavior 
- **No content flicker** - protected page content never appears
- **Blank/empty screen** - only loading state or nothing shown
- **Smooth redirect** - immediate redirect to login page
- **Clean UX** - no visual glitching or flash

### Verification Results
- Direct navigation to protected routes shows NO flicker
- Auth state checked BEFORE rendering content
- Smooth redirect to login WITHOUT UI flash
- No console errors related to auth guards

**Pattern Used:** Sync External Store with conditional rendering based on state

**Key Insight:** In Next.js with client components, hydration-safe state checking requires:
1. Use `useSyncExternalStore` for storage state synchronization
2. Wait for `ready` state before rendering protected content
3. Return `null` during auth check to prevent flicker

**Resources:**
- [React useSyncExternalStore](https://react.dev/reference/react/useSyncExternalStore)
- [Next.js Auth Patterns](https://nextjs.org/docs/app/building-your-application/authentication)
- Original Issue: #1088 - Auth guard flicker shows protected page before redirect

## Files Changed

| File | Changes |
|------|---------|
| `web/src/components/layouts/auth-guard.tsx` | Added `ready` state check, return `null` if not ready |
| `web/src/components/layouts/role-guard.tsx` | Added `ready` state check, return `null` if not ready |

### DIFF
--- a/web/src/components/layouts/auth-guard.tsx
+++ b/web/src/components/layouts/auth-guard.tsx
@@ -6,9 +6,10 @@
 import { useAuthGuard, useOptionalAuth } from "@/hooks/use-auth";
 
 export function AuthGuard({ children }: { children: React.ReactNode }) {
-  useAuthGuard();
-  // Render children immediately to prevent hydration mismatch
-  // The useAuthGuard hook handles redirects via side effects
+  const { ready } = useAuthGuard();
+  // Block rendering until auth state is confirmed to prevent flicker
+  // of protected content before redirect
+  if (!ready) return <div className="flex h-screen w-full items-center justify-center" />;
   return <>{children}</>;
 }
 
--- a/web/src/components/layouts/role-guard.tsx
+++ b/web/src/components/layouts/role-guard.tsx
@@ -7,8 +7,9 @@
 import { useRoleGuard, type Role } from "@/hooks/use-role-guard";
 
 export function RoleGuard({ minRole, children }: { minRole: Role; children: React.ReactNode }) {
-  useRoleGuard(minRole);
-  // Render children immediately to prevent hydration mismatch
-  // The useRoleGuard hook handles redirects via side effects
+  const { ready } = useRoleGuard(minRole);
+  // Block rendering until role is confirmed to prevent flicker
+  // of protected content before redirect
+  if (!ready) return <div className="flex h-screen w-full items-center justify-center" />;
   return <>{children}</>;
 }


Haz3-jolt:
Rebased onto main and made a small adjustment to the loading state: replaced return null with an empty full-height container (<div className="flex h-screen w-full items-center justify-center" />) in both AuthGuard and RoleGuard. This avoids a potential hydration mismatch (server and client both render the same empty div when auth is unresolved) while still blocking protected content from flashing before the redirect fires.

Haz3-jolt (approved):
Thanks so much!!!!

PR #1156, opened 2026-05-20
fix: scope reconcile dedup by user

tsitu0:
## Purpose / Description
Fix a session reconciliation edge case where duplicate checks were not scoped to the current user.

The reconcile endpoint already checks that the session belongs to the authenticated user before writing enrichment data. However, the "already reconciled" check only looked at `session_id` or `subagent_id`.

That means if two users ever had the same session ID or subagent ID, one user's reconciliation record could cause the other user's reconciliation to be skipped.

## Fixes
* No linked issue

## Approach
Added `user_id` to reconcile enrichment log attributes.

Updated the duplicate check to require both:

```text
same session/subagent ID
same user ID
```
This keeps reconciliation idempotent for each user without allowing one user's reconcile state to affect another user's session.

## How Has This Been Tested?

Ran the focused reconciliation tests:
```bash
uv run --project observal-server --with pytest --with pytest-asyncio pytest tests/test_reconcile_subagent.py -q
```
Result:
```bash
15 passed, 2 skipped
```
Also ran:
```bash
uv run ruff check observal-server/api/routes/reconcile.py tests/test_reconcile_subagent.py
git diff --check
```
Result:
```bash
All checks passed
```

This follows the same idea as the existing ownership check: session data should be scoped to the authenticated user. The endpoint already checks ownership before writing; this change also scopes the duplicate check by user.

## Checklist

- [x] UI changes: include screenshots of all affected screens (N/A, backend only)


tests/test_reconcile_subagent.py
```diff
@@ -374,6 +380,83 @@ async def test_reconcile_endpoint_subagent_dedup_uses_subagent_id_not_session_id
...
+    assert "LogAttributes['subagent_id'] = {key:String}" in dedup_sql
+    assert "LogAttributes['user_id'] = {uid:String}" in dedup_sql
+    assert dedup_params["param_key"] == "agent-already-done"
+    assert dedup_params["param_uid"] == str(mock_user.id)
+
+
+@pytest.mark.asyncio
+async def test_reconcile_endpoint_dedup_is_scoped_by_user():
+    """Top-level session dedup includes user_id so users cannot affect each other."""
+    from fastapi import FastAPI
```
ShaanNarendran: can move the imports to the top of the file, otherwise just fix the conflicts and LGTM

PR #1225, opened 2026-05-25
Fix/agent rbac

harishankar0301:
Removes the unintended org-wide visibility bypass for private agents in `list_agents` and `get_effective_agent_permission`. Org members must now explicitly have team access via SSO groups to view or edit private agents. Public agent scoping is retained.

## Purpose / Description
Enforce Group based Access and Scoped Visibility for Private Agents

## Problem
Previously, the agent access control logic contained an unintended bypass that leaked private agents to all users within the same organization. Specifically:
1. In `api/routes/agent/crud.py` (`list_agents`), the SQL query builder added an unconditional `OR` clause allowing visibility of any agent owned by the current user's organization, entirely bypassing `AgentVisibility.private` settings.
2. In `api/deps.py` (`get_effective_agent_permission`), a similar early return automatically granted `"view"` access to anyone in the same organization. This not only broke privacy but also prevented users from getting `"edit"` permissions if they were legitimately assigned to an editor group (since the code returned `"view"` before evaluating `team_accesses`).

## Approach
This PR removes the organization-wide bypass for private agents, ensuring that `AgentVisibility.private` and `AgentTeamAccess` act as the definitive sources of truth for RBAC enforcement. 
**Changes Made:**
- **`api/routes/agent/crud.py`:**
  - Modified the `visibility_filter` to ensure that being in the same organization only grants registry visibility if the agent is explicitly marked as `AgentVisibility.public` (`Agent.visibility == AgentVisibility.public`).
- **`api/deps.py`:**
  - Removed the unconditional early `"view"` return for matching `org_id`.
  - Shifted the evaluation order: `team_accesses` are now correctly processed *first*, allowing group-based `"edit"` permissions to take precedence.
  - Re-introduced the `org_id` visibility check strictly as a fallback for `AgentVisibility.public` agents, maintaining the standard public org-scoping without compromising private agent security.

- **Private Agents:** Are now fully hidden and inaccessible to other users in the organization unless they are the owner, an admin, or belong to an SSO group defined in the agent's `team_accesses`.
- **Public Agents:** Behave as expected, remaining viewable to anyone within the same organization. 
- **Team Access:** Users granted `"edit"` access via SSO groups are correctly evaluated and assigned proper permissions, rather than being downgraded to `"view"`.

## How Has This Been Tested?

- Existing SAML tests pass (make test)
- Tested against Okta developer tenant
- Added tests to check enforcement of the RBAC control


harishankar0301:
Need to add tests to enforce this controls.

Haz3-jolt:
Removed agent visibility

PR #1260, opened 2026-05-28
fix: insights session count, error reporting, and adaptive metrics

HemalathaMadeswaran18:
## Purpose / Description

Agent Insights was showing 0 sessions and empty reports due to multiple issues in the data pipeline:
1. ClickHouse queries missing `FINAL` keyword caused unmerged `agent_id` values to be invisible
2. Poisoned `first_event_time` (epoch zero) excluded all sessions from the time filter
3. Bedrock API errors were silently swallowed, producing empty "completed" reports
4. Admin setting changes were not reflected until server restart (stale in-memory cache)
5. Kiro session tool usage (file writes, git commits) was not being extracted
6. Frontend showed misleading zeros and empty cost sections for IDEs that don't report token data

## Fixes

* Fixes insights showing "0 sessions available" for agents with valid sessions
* Fixes empty insight reports when AWS credentials are invalid/expired
* Fixes stale settings cache requiring server restart after admin changes
* Fixes Kiro sessions not reporting lines added/files modified/git stats
* Fixes misleading "$0.00 cost" display for Kiro-only agents

### Backend (observal-server)

- **`api/routes/insights.py`**: Added `FINAL` to `session_stats_agg` queries; switched time filter from `first_event_time` to `last_event_time`; replaced `get_sync()` with async `ds.get()` in `/status` endpoint; added STS-based AWS credential validation with descriptive error messages
- **`api/routes/admin/enterprise_settings.py`**: Added `ds.invalidate(key)` + `ds.refresh_sync_cache()` after setting upsert/delete so changes take effect immediately
- **`services/insights/__init__.py`**: `_call_bedrock` now raises `RuntimeError` with descriptive messages (invalid credentials, access denied, model not available, expired token) instead of returning `{}`

### Insights Engine (ee/observal_insights)

- **`sections.py`**: `_call_section` re-raises `RuntimeError` so credential/config errors propagate to the report level (sets `status: failed` with `error_message`)
- **`session_meta_extractor.py`**: Added `FINAL` to session ID query; added file/lines/git extraction from Kiro `toolUse` entries (parses `write`, `edit`, `bash` tool inputs for file paths, content, and git commands)

### Frontend (web)

- **`insights/page.tsx`**: Show loading indicator instead of "0" while session count is fetching; don't disable Generate button during loading
- **`insights/[reportId]/page.tsx`**: Hide metric cards (Tokens In/Out, Cache, Cost, Lines, Git, Errors, Interruptions) when value is 0; hide "Cost & Token Efficiency" narrative section when no token/cost data exists
- **`use-insights-api.ts`**: Set `staleTime: 0` on status query so it always fetches fresh on page load

## How Has This Been Tested?

1. Verified `session_stats_agg` returns correct `agent_id` with `FINAL` via direct ClickHouse queries
2. Confirmed `/session-count` endpoint returns correct count after fix
3. Tested AWS credential validation: invalid key -> "AWS access key is invalid", wrong secret -> "AWS secret key is invalid"
4. Verified Bedrock throttling errors now set report `status: failed` with descriptive `error_message`
5. Confirmed admin setting changes take effect immediately without container restart
6. Verified Kiro tool use extraction: `write` tool -> lines_added + files_modified populated
7. Confirmed metric cards and Cost section hidden when values are 0

## Learning

- ClickHouse `AggregatingMergeTree` with non-key plain columns (`agent_id`) requires `FINAL` at read time to see merged values — without it, unmerged parts may return stale/empty values
- `SimpleAggregateFunction(min, DateTime64)` can be poisoned by epoch-zero timestamps from the first batch insert, making time-range filters exclude all data
- `get_sync()` in dynamic_settings uses an in-memory dict populated once at startup — writes must explicitly call `refresh_sync_cache()` to update it
- Kiro JSONL format does not include token/cost metadata — only conversation content and tool use. Metrics that depend on API-level data (tokens, cost) will always be 0 for Kiro sessions

## Checklist

### DIFF
--- a/ee/observal_insights/generator.py
+++ b/ee/observal_insights/generator.py
@@ -190,6 +190,7 @@ async def _run_pipeline(
             "subagent_sessions": agg.get("sessions_using_subagent", 0),
             "mcp_sessions": agg.get("sessions_using_mcp", 0),
             "total_cost_usd": round(agg.get("total_cost", 0), 2),
+            "total_credits": round(agg.get("total_credits", 0), 4),
             "total_input_tokens": agg.get("total_input_tokens", 0),
             "total_output_tokens": agg.get("total_output_tokens", 0),
             "total_cache_read_tokens": agg.get("total_cache_read_tokens", 0),
@@ -198,6 +199,9 @@ async def _run_pipeline(
             "top_languages": agg.get("top_languages", [])[:10],
             "tool_error_categories": agg.get("tool_error_categories", {}),
             "projects": agg.get("projects", {}),
+            "ides": agg.get("ides", []),
+            "sessions_with_tokens": agg.get("sessions_with_tokens", 0),
+            "sessions_with_credits": agg.get("sessions_with_credits", 0),
         },
         "overview": {
             "total_sessions": agg.get("total_sessions", 0),
--- a/ee/observal_insights/generator.py
+++ b/ee/observal_insights/generator.py
@@ -147,6 +147,111 @@ async def _run_pipeline(
 
     logger.info("insight_facets_extracted", count=len(all_facets))
 
+    # ── Step 3b: Model efficiency analysis ─────────────────────────────────
+    # Cross-reference per-session model usage with facets to detect waste.
+    # Build session_id -> facets mapping from the extraction batch.
+    facets_by_session: dict[str, dict] = {}
+    if transcripts and all_facets:
+        session_ids_with_transcripts = list(transcripts.keys())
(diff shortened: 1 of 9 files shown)


HemalathaMadeswaran18:
Screenshot 2026-05-28 at 7 25 08 PM
 this is now not shown, if zero

HemalathaMadeswaran18:
image
this is now removed if null /0

HemalathaMadeswaran18:
image

Haz3-jolt (approved):

PR #1298, opened 2026-05-30
feat(security): make sensitive settings write-once and retractable

ShaanNarendran:
## Purpose / Description

Sensitive API keys and secrets stored in super admin settings (AWS credentials, SAML certificates, encryption passwords) were previously returned as plaintext in API responses. The frontend masked them visually, but the raw values were exposed in network traffic. This change enforces server-side write-once semantics and adds a revocation mechanism.

## Fixes

* Fixes the security gap where sensitive settings could be read back via the API after initial entry

## Approach

**Server-side (enterprise_settings.py):**
- GET /admin/settings and GET /admin/settings/{key} now return `**REDACTED**` (from `services.secrets_redactor.REDACTED`) for any key in `SENSITIVE_KEYS`. The plaintext is never sent over the wire after the initial save.
- PUT /admin/settings/{key} encrypts sensitive values with Fernet before DB storage. The response confirms save with the redacted constant.
- New POST /admin/settings/{key}/revoke endpoint permanently deletes a sensitive key, restricted to `SENSITIVE_KEYS` only, with a CRITICAL-severity audit event.

**Schema (schemas/admin.py):**
- `EnterpriseConfigResponse` extended with `is_sensitive: bool` and `is_set: bool` fields so the frontend can render appropriate UI without needing client-side secret knowledge.

**Frontend (settings.tsx, api.ts, types/admin.ts):**
- `SENSITIVE_KEYS` expanded to match the server set (added `insights.aws_*`).
- Removed client-side `maskValue()` function; relies on server-provided `is_sensitive`/`is_set`.
- Sensitive keys show `**REDACTED**` when set, with a revoke (trash) button instead of the generic clear.
- Edit always starts with empty input for sensitive keys (cannot pre-fill, value is not available).
- Added `admin.revokeSetting()` API client method.

## How Has This Been Tested?

- `uv run pytest tests/test_enterprise.py` passes (6/6)
- `ruff check` passes on all modified Python files
- TypeScript compiles clean (`tsc --noEmit --skipLibCheck`)
- Manual review of the API response contract: sensitive keys return `**REDACTED**` in list/get, and the PUT response also returns redacted

## Checklist

### DIFF
--- a/observal-server/api/routes/admin/enterprise_settings.py
+++ b/observal-server/api/routes/admin/enterprise_settings.py
@@ -14,7 +14,8 @@ from api.deps import get_db, require_role
 from config import HAS_LICENSE, settings
 from models.enterprise_config import EnterpriseConfig
 from models.user import User, UserRole
-from schemas.admin import EnterpriseConfigResponse, EnterpriseConfigUpdate
+from schemas.admin import EnterpriseConfigResponse, EnterpriseConfigUpdate, SettingRevokedResponse
+from services.secrets_redactor import REDACTED
 from services.security_events import EventType, SecurityEvent, Severity, emit_security_event
 
 from ._router import router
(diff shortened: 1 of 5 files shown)


Haz3-jolt (changes requested):
LGTM

PR #1318, opened 2026-05-31
fix(security): redact MCP clone tokens

### LINKED ISSUE #1317: MCP validation should redact clone tokens from persisted errors
### Checked for duplicates?

MCP validation builds authenticated clone URLs when `GIT_CLONE_TOKEN` is configured.

If the Git clone operation fails, the validation path stores the raw clone exception in `McpValidationResult.details`.

Some Git clone errors can include the authenticated URL, which may contain the configured token.

### Expected behaviour

Validation errors should never persist secret material.

Clone failure details should redact the configured `GIT_CLONE_TOKEN` before saving the error, and should also pass through the existing secret redaction helper for defense in depth.

### Support bundle

Not applicable. This is a static backend hardening issue.

This is related to MCP validation error handling and can be covered with a focused regression test.

tsitu0:
## Purpose / Description

MCP validation supports private repositories by injecting `GIT_CLONE_TOKEN` into the Git clone URL before calling Git.

If the clone fails, the validation path currently stores the raw exception text in `McpValidationResult.details`. Git clone errors can include the URL that was attempted, which means an authenticated clone URL may be persisted with the token included.

That creates a secret exposure risk in validation results.

## Fixes

Fixes #1317 

## Approach

This change sanitizes clone failure messages before they are saved.

It adds a small helper that:

1. Replaces the configured `GIT_CLONE_TOKEN` with the project’s redaction marker.
2. Passes the resulting message through the existing `redact_secrets()` helper for defense in depth.
3. Uses the sanitized message when writing `McpValidationResult.details`.

This keeps the validation error useful while preventing the clone token from being stored.

A regression test was added for the specific failure mode: a clone exception containing an authenticated Git URL.

## How Has This Been Tested?

Ran the targeted backend regression test file:

```bash
uv run --project observal-server pytest tests/test_audit2_pr1_ingest_bounds.py -q
```

Result:

```text
22 passed in 0.59s
```
Commit hooks also passed when creating the commit.

I followed the existing secret-redaction pattern already used elsewhere in the backend, especially the recent hardening around ingest and GraphQL read paths.

## Checklist
- [x] You have commented your code, particularly in hard-to-understand areas.
- [x] You have performed a self-review of your own code.
- [x] UI changes: not applicable; backend-only change.

### DIFF
--- a/observal-server/services/mcp_validator.py
+++ b/observal-server/services/mcp_validator.py
@@ -2,6 +2,7 @@
 # SPDX-FileCopyrightText: 2026 Hari Srinivasan <harisrini21@gmail.com>
 # SPDX-FileCopyrightText: 2026 Kaushik Kumar <kaushikrjpm10@gmail.com>
 # SPDX-FileCopyrightText: 2026 Shaan Narendran <shaannaren06@gmail.com>
+# SPDX-FileCopyrightText: 2026 tsitu0 <tomsitu0102@gmail.com>
 # SPDX-License-Identifier: AGPL-3.0-only
 
 import ast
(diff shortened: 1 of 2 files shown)


Haz3-jolt (approved):
Thanks!

PR #1327, opened 2026-06-01
fix(auth): use platform-aware browser open for device flow

shreem26:
## Purpose

Fix `observal auth login` printing `gio: Operation not supported` error on WSL when opening the browser for device authorization.

## Approach

**Problem:** `webbrowser.open()` on WSL calls `gio` (a Linux desktop tool) which fails with "Operation not supported" since WSL has no native browser. The error message is printed to the terminal, confusing users.

**Fix:** Replace `webbrowser.open()` with platform-aware subprocess calls:
- **macOS:** `open <url>`
- **Linux (WSL):** `powershell.exe Start-Process <url>` — opens in the Windows default browser
- **Linux (native):** `xdg-open <url>`
- **Windows:** falls back to `webbrowser.open()`

WSL detection uses `wslpath -w /` (returns 0 only on WSL). All subprocess calls suppress stderr/stdout to avoid terminal pollution.

## How has this been tested

- Manual testing on WSL: option [2] opens browser via `powershell.exe` without `gio` errors
- `observal_cli/tests/test_cmd_auth.py` — 10/10 passing

### DIFF
--- a/observal_cli/cmd_auth.py
+++ b/observal_cli/cmd_auth.py
@@ -553,8 +553,31 @@ def _do_device_flow_login(server_url: str):
 
     # Try to open browser automatically
     try:
-        webbrowser.open(verification_uri_complete)
-        rprint("[dim]Browser opened automatically.[/dim]")
+        import platform
+        import subprocess as _sp
+
+        _opened = False
+        _sys = platform.system()
+        if _sys == "Darwin":
+            _sp.Popen(["open", verification_uri_complete], stderr=_sp.DEVNULL, stdout=_sp.DEVNULL)
+            _opened = True
+        elif _sys == "Linux":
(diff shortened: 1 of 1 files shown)


Haz3-jolt:
@shreem26 No need for a separate SSO option if you're taken to the same login screen anyway

shreem26:
@Haz3-jolt okayyy, what about fix 2?

Haz3-jolt:
Fix 2 is good stuff!

shreem26:
@Haz3-jolt should I create a seperate pr then implementing fix 2 alone and close this one?

Haz3-jolt:
Nope, just update the PR to only implement fix2.

shreem26:
@Haz3-jolt heyy i updated the pr

PR #1339, opened 2026-06-03
fix(device-auth): resolve verification_uri to actual deployment URL

ShaanNarendran:
## Purpose / Description

The device authorization flow (`observal auth login --sso`) always showed `http://localhost/device` as the verification URL, even on deployed instances. Users on remote servers saw a localhost URL that could not be reached from their browser.

## Fixes

* Fixes the device auth verification_uri showing `http://localhost/device` instead of the actual deployment URL

## Approach

The root cause was in `_resolve_frontend_url()` in `device_auth.py`. The function loaded the DEFAULTS value (`"http://localhost"`) from the dynamic settings cache, then compared it against a hardcoded `"http://localhost:3000"`. Since these strings never match, the function treated localhost as an intentionally configured URL and returned it immediately, never attempting header inference.

The fix rewrites `_resolve_frontend_url` with a proper resolution chain:

1. `deployment.frontend_url` if explicitly set to a non-localhost value
2. `deployment.public_url` hostname as a fallback (already used by other routes)
3. Request `Host` / `x-forwarded-*` headers (works behind any reverse proxy, not just HTTPS)
4. `http://localhost` only as a true local-dev fallback

A new `_is_localhost_url` helper detects unconfigured state regardless of port variations (`localhost`, `localhost:3000`, `127.0.0.1`, etc.)

## How Has This Been Tested?

Added 7 unit tests for `_resolve_frontend_url` covering:
- Explicitly configured frontend URL (non-localhost)
- Trailing slash stripping
- Fallback to `deployment.public_url` when frontend is localhost
- Inference from `x-forwarded-proto` + `Host` headers
- Inference from `Host` header alone (no forwarded proto)
- Localhost fallback when no useful context exists
- `localhost:3000` also treated as unconfigured

All 18 non-pre-existing tests pass.

## Learning

The `derive_endpoints()` function in `config.py` already handled this correctly for the `/config/endpoints` route. This fix aligns the device auth URL resolution with the same pattern, prioritizing `deployment.public_url` and request headers over localhost defaults.

## Checklist

### DIFF
--- a/observal-server/api/routes/device_auth.py
+++ b/observal-server/api/routes/device_auth.py
@@ -14,6 +14,7 @@ The flow:
 import json
 import secrets
 import time
+from urllib.parse import urlparse
 
 from fastapi import APIRouter, Depends, HTTPException, Request
 from fastapi.responses import JSONResponse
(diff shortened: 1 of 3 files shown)


shreem26:
looks good to me

PR #1371, opened 2026-06-06
test(version-check): use UTC clock for short-TTL cache timestamp

### LINKED ISSUE #1370: fix(tests): test_uses_short_ttl_cache fails on non-UTC machines
## Bug

  `test_uses_short_ttl_cache` in `tests/test_version_check.py` fails on any machine not set to UTC.

  ## Root Cause

  Line 370 uses `time.strftime("%Y-%m-%dT%H:%M:%S+00:00")` which gets **local time** but hardcodes a UTC
  offset (`+00:00`). The production code (`_should_check`) then parses this with `datetime.fromisoformat()`
  and compares against `time.time()` (actual UTC). The mismatch equals the machine's UTC offset.

  - UTC-4 (New York): test thinks cache is 4 hours old → network fires → assertion fails
  - UTC+5:30 (India): timestamp lands in the future → clock-skew guard triggers → assertion fails

  ## Reproduce

  TZ=America/New_York make test
  test_uses_short_ttl_cache FAILS

  ## Fix

  Replace line 370 in tests/test_version_check.py:

  # Before
  "last_checked": time.strftime("%Y-%m-%dT%H:%M:%S+00:00"),

  # After
  "last_checked": datetime.now(UTC).strftime("%Y-%m-%dT%H:%M:%S+00:00"),

  Add from datetime import datetime, UTC at the top of the test method.

  # Why CI passes

  GitHub Actions runners use UTC, so the local time and UTC happen to match.

ign-ite:
## Summary
  Fixes the flaky `test_uses_short_ttl_cache` failure on machines in timezones
  ahead of UTC (e.g. IST).

  The test stamped `last_checked` with `time.strftime("...+00:00")`, which uses
  the **local** clock but labels it UTC. In `_should_check`
  (`observal_cli/version_check.py:199`) the clock-skew guard then sees a
  "future" timestamp, treats the cache as stale, and forces a network call —
  breaking `assert network_called["hit"] is False`.

  Using `time.gmtime()` makes the `+00:00` label match the actual instant, so
  the test is correct in every timezone (matches production, which writes
  `datetime.now(UTC).isoformat()`).

  ## Test
  49/49 pass in `tests/test_version_check.py` under `TZ=Asia/Kolkata`.

  Closes #1370

### DIFF
--- a/tests/test_version_check.py
+++ b/tests/test_version_check.py
@@ -367,7 +367,7 @@ class TestCheckVersionCompatibility:
         fresh_cache = {
             "server_version": "1.0.5",
             "source": "server",
-            "last_checked": time.strftime("%Y-%m-%dT%H:%M:%S+00:00"),
+            "last_checked": time.strftime("%Y-%m-%dT%H:%M:%S+00:00", time.gmtime()),
         }
         monkeypatch.setattr(version_check, "_read_cache", lambda: fresh_cache)
         network_called = {"hit": False}


Kaushik-Kumar-CEG:
LGTM!
@Haz3-jolt can be merged

Haz3-jolt (approved):
Clean fix. Using `time.gmtime()` ensures the timestamp is always UTC regardless of the runner's timezone, preventing flaky test failures. LGTM.

Haz3-jolt:
@ign-ite use PR template

ign-ite:
Sorry about that! Will use the PR Template next time!

PR #1411, opened 2026-06-10
feat(auth): add Google sign-in

Apoorvgarg-creator:
## Purpose / Description

Adds **Google** as a first-class sign-in option for self-hosted Observal. Admins drop in their OAuth client ID and secret (and, if they want, an email-domain allowlist) and a real "Sign in with Google" button shows up on the login page. It coexists with the existing generic OIDC slot, so an org running Okta or Keycloak can keep that and offer Google on the same screen.

The motivation: the generic OIDC slot already technically supports Google (point it at `accounts.google.com/.well-known/openid-configuration` and it works), but the button just says "Sign in with SSO". That's fine for one internal IdP and confusing for everyone else. Making Google a named provider gets us a branded button, lets it sit alongside the custom OIDC, and gives us a clean place to plug in Google-specific checks like `email_verified` and `hd`-style domain restrictions.

## Fixes

N/A — new feature, no linked issue.

## Approach

**Backend.** I added a second Authlib client called `"google"` and hardcoded the discovery URL server-side, so admins only ever set client ID + secret. Two new routes — `/api/v1/auth/oauth/google/login` and `/api/v1/auth/oauth/google/callback` — handle the redirect dance.

While in there I pulled the duplicated bits out of the existing `/oauth/callback` into two helpers (`_provision_sso_user` and `_complete_sso_login`) so the Google route reuses them instead of copy-pasting. The generic OIDC callback now also records `auth_provider` and `sso_subject_id` on the user row, which it wasn't doing before — small bonus correctness fix.

One subtle ordering thing: the `SSO_SUCCESS` audit event now fires *after* the DB commit. Before, a failed commit would still log a success. That bug was inherited from the original code; the helper extraction made it easy to fix.

The Google route adds two checks the generic OIDC doesn't have: it requires Google's `email_verified` claim to be true, and if `GOOGLE_OAUTH_ALLOWED_DOMAINS` is set (comma-separated list like `acme.com,acme.io`), it rejects anything outside the list with a 403 + audit event. Both are intentionally Google-only — generic OIDC providers handle verification their own way and you wouldn't want shared semantics there.

**Frontend.** `/api/v1/config/public` exposes a new `google_sso_enabled` flag. The `useDeploymentConfig` hook surfaces it. The login page renders an extra "Sign in with Google" button above the existing SSO/SAML ones. The button uses the existing `Button` component with `variant="outline"` so it inherits the theme tokens consistently. The "G" itself is a tiny inline SVG component I added — couldn't find an existing Google icon in the UI kit, and `lucide-react` doesn't ship one. Verified by grep before adding.

**Self-hosting.** `.env.example` documents the three new variables. GCP Terraform threads the client ID and secret through Secret Manager and exposes the allowlist as a plain Cloud Run env var. AWS-EC2 uses an `env_overrides` map that already accepts arbitrary vars, so the change there is documentation plus an example in the README. The docker-compose file pulls from `.env` directly so it picks up the new vars without changes.

One subtle gotcha I worked around: the GCP `secrets.tf` does `coalesce(each.value, " ")` because Secret Manager rejects empty values, which means unset Google secrets land in Cloud Run as the literal string `" "` (one space). Python sees that as truthy. So I compute `GOOGLE_SSO_ENABLED` once in `config.py` with `.strip()` and import it from there; both the auth registration and `/config/public` use the same value, no drift. The Terraform side could be fixed cleaner with `dynamic env` blocks, but that's an orthogonal refactor and would touch all the optional secrets in the same file.

**Docs.** `docs/self-hosting/authentication.md` got a new "Google OAuth (first-class provider)" subsection with the GCP Console steps, redirect URI, allowlist behavior, and first-login notes.

**No DB migration.** `User.auth_provider` and `User.sso_subject_id` columns already exist from the SSO work.

## How Has This Been Tested?

```bash
pytest tests/test_google_oauth.py -v
```

The test file covers:

- Routes 500 cleanly when Google OAuth env vars aren't set.
- Callback rejects missing userinfo, missing email claim, `email_verified=false` (400).
- Out-of-allowlist domains are rejected with 403 + audit event.
- Allowlisted domain reaches `_provision_sso_user` with `provider="google"` and the Google `sub` recorded as `sso_subject_id`.
- `_provision_sso_user` direct tests: creates new users, returns existing ones unchanged, upgrades a previously-local user on first SSO sign-in without overwriting an existing `sso_subject_id`.
- `_complete_sso_login` happy path: commit fires before the success event, code lands in Redis, redirect points at `/login?code=...`.
- The refactored generic `/oauth/callback` calls `_provision_sso_user` with `provider="oidc"` (regression check on the helper extraction).
- `_is_safe_next` rejects unsafe paths (`//evil`, `https://...`, backslashes) and accepts plain local paths.
- `_parse_allowed_domains` lowercases, strips, and ignores empty segments.
- `/config/public` returns `google_sso_enabled: true` when the constant is set, `false` otherwise.

Manual verification I'd like a reviewer to run before merging:

- [ ] Spin up local docker-compose with `GOOGLE_OAUTH_CLIENT_ID/SECRET` filled in, hit `/login`, click the Google button, end up signed in on `/`.
- [ ] Check the new user row has `auth_provider='google'` and a numeric `sso_subject_id`.
- [ ] Set `GOOGLE_OAUTH_ALLOWED_DOMAINS=somedomain.com`, restart the API container, confirm an out-of-domain account is rejected with the 403 page and an `SSO_FAILURE` event lands in the audit log.
- [ ] Leave both Google vars unset, confirm the login page is identical to today (no Google button, no console errors, password login still works).

## Learning (optional)

The redirect URI Google enforces (`{FRONTEND_URL}/api/v1/auth/oauth/google/callback`) is a one-way door — once admins paste it into the Google Cloud Console, changing the path means going back into the console. So the path naming got more deliberation than usual. Nested under `oauth/` keeps all the OAuth flows grouped, and `/google/` leaves the door open for `/microsoft/`, `/github/` later without renaming.

Authlib's discovery URL fetch is lazy (first redirect/callback, not at `oauth.register`), so adding Google doesn't add a startup network call.

## Checklist

- [x] Descriptive commit messages with short titles (each under 30 chars per project style).
- [x] Self-reviewed.
- [x] No new external libraries introduced — Google "G" SVG is an inline component I wrote.
- [ ] UI screenshots: pending. Will attach to the PR once the demo deploy is up.

---

**PS** — me sit by fire, me look at screen. Big yellow circle on login button now. Press button, browser go somewhere, come back. Now me inside app. Magic rectangle work. Good. Me eat. 🦴


Haz3-jolt:
No merge commits please.

Apoorvgarg-creator:
> No merge commits please.

Screenshot 2026-06-10 at 9 39 55 PM

You mean this ?

I don't see any change as well. I don't know why I got this.

PR #1427, opened 2026-06-12
feat(ops): add insights CLI pretty print commands

shreem26:
## Purpose / Description

Add `observal ops insights` subcommand group to the CLI, enabling users to list, generate, and pretty-print Agent Insight reports directly from the terminal without opening the web UI.

## Approach

- Added `observal_cli/cmd_insights.py` with three commands:
  - `observal ops insights list <agent>` — tabular list of reports with status, period, session count
  - `observal ops insights generate <agent>` — trigger a new report, with a pre-check that shows setup instructions if the LLM model is not configured
  - `observal ops insights show <report-id>` — full pretty-printed narrative using Rich panels and tables, one renderer per section (`at_a_glance`, `what_works`, `friction_analysis`, `suggestions`, `on_the_horizon`, etc.). Supports `--section <name>` to render a single section and `--output json` for machine-readable output.
- Wired `insights_app` into `ops_app` in `main.py`
- Fixed `version_check.py` bug: `auto_update` stored as string `"false"` was not being parsed correctly (non-empty string is truthy in Python). Now handles both string and bool values.
- Changed CLI-ahead-of-server from a hard block to a warning, so local development with unreleased CLI changes is not blocked.

## How Has This Been Tested?

1. `make rebuild` to start the local Docker stack
2. `uv tool install --editable . --force --reinstall` to install the editable CLI
3. Verified `observal ops insights --help` shows `list`, `show`, `generate`
4. Verified `observal ops insights generate myagent` shows setup instructions when `insights.model_sections` is not configured
5. Verified `observal ops insights list myagent` returns a table once a report exists
6. Verified `--output json` passes raw data through unchanged
7. Verified `auto_update` string fix stops the binary from being silently overwritten

## Learning (optional)

- `uv tool install --editable .` builds from local source but `auto_update_if_needed()` was silently replacing the binary with the latest remote release on every command. The root cause was a type mismatch: `observal config set` stores values as JSON strings, but the check used Python truthiness on the string `"false"`.
- Server version enforcement hard-blocked CLI-ahead-of-server, which breaks local development. Changed to a warning since the server being behind is not a CLI problem.

## Checklist

## AI Assistance

- [x] Yes (Kiro)

### DIFF
--- a/observal-server/api/routes/agent/__init__.py
+++ b/observal-server/api/routes/agent/__init__.py
@@ -4,7 +4,7 @@
 """Agent routes package. Sub-modules register routes on the shared router."""
 
 # Import sub-modules so they register their routes on the shared router.
-from . import crud, draft, install  # noqa: F401
+from . import crud, draft, insights, install  # noqa: F401
 from ._router import router  # noqa: F401
 from .helpers import _load_agent, _resolve_component_names  # noqa: F401
 from .install import install_agent  # noqa: F401
--- a/observal-server/api/routes/agent/insights.py
+++ b/observal-server/api/routes/agent/insights.py
@@ -1,0 +1,114 @@
+# SPDX-FileCopyrightText: 2026 Hari Srinivasan <harisrini21@gmail.com>
+# SPDX-License-Identifier: AGPL-3.0-only
+
+"""Agent-scoped insight report routes."""
+
+from fastapi import Depends, HTTPException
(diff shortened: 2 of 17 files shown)


ShaanNarendran:
LGTM, needs a manual test (my api key is invalid)

ShaanNarendran (approved):

Haz3-jolt (approved):

PR #1446, opened 2026-06-16
fix(insights): include untagged sessions in session count

HemalathaMadeswaran18:
## Purpose / Description
Insights session count endpoint returns 0 even when sessions exist for an agent, because sessions ingested via hooks do not carry `agent_version` (resolves to empty string in ClickHouse). The version filter was excluding all untagged sessions.

## Fixes
* Fixes the "0 sessions available" display on the Insights tab when sessions exist

## Approach
Modified the version filter in `_count_insight_sessions` (insights.py) and `_count_agent_sessions` (batch.py) to also match `agent_version = ''` (untagged sessions). The fallback query on `session_events` additionally matches `agent_version IS NULL`.

This does not affect `version_impact.py` or `session_meta_extractor.py` — those either pass `None` (matches all) or intentionally require strict version grouping for comparison.

## How Has This Been Tested?

- Ruff lint passes on both modified files
- AST parse verification passes
- Local Docker stack tested with a dummy agent (`test-dummy-agent`) and session ingest
- Reviewed version_impact and session_meta_extractor queries to confirm no regression

ClickHouse `session_stats_agg` MV coalesces NULL agent_version to empty string. Most sessions ingested via hooks only carry agent_id (name or UUID), not agent_version, because `_resolve_agent()` returns version only from lockfile lookups.

## Checklist

## AI Assistance
- [x] Yes(Please Specify the tool): Kiro CLI

### DIFF
--- a/observal-server/api/routes/insights.py
+++ b/observal-server/api/routes/insights.py
@@ -139,7 +139,9 @@ async def _count_insight_sessions(
         "param_t_end": period_end.strftime("%Y-%m-%d %H:%M:%S"),
     }
     if agent_version:
-        base_where += "AND agent_version = {agent_version:String} "
+        # Include sessions with matching version OR untagged sessions (agent_version = '')
+        # because most sessions are ingested without explicit version tagging.
+        base_where += "AND (agent_version = {agent_version:String} OR agent_version = '') "
         params["param_agent_version"] = agent_version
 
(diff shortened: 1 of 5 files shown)


HemalathaMadeswaran18:
Screenshot 2026-06-16 at 7 42 03 PM

PR #1257, opened 2026-05-27
fix: Kiro credits, traces UX, auth, insights, and misc fixes

Haz3-jolt:
## Purpose / Description

Batch of fixes addressing Kiro session handling, traces page UX, auth flow issues, insight engine Kiro support, and misc UI polish.

## Fixes

* Kiro credits not showing on single-prompt sessions (race condition on stop hook)
* Duration and active status broken for Kiro sessions (sentinel timestamp poisoning MV)
* Traces not chronologically sorted by default
* Login autofilling http://localhost:80 instead of leaving blank
* Middle-click opening new tab requiring re-login (sessionStorage vs localStorage)
* CLI login options redundant (SSO vs browser unified)
* Review diff dialog animation glitch and jank on heavy content
* Approved badge blinking, agent status not updating after approval
* Hardcoded platform filter in traces replaced with parameterized search
* Sessions capped at 100 with no pagination (now 50/page with Prev/Next)
* Active sessions threshold too generous (30 min to 5 min)
* Device auth polling rate-limited too aggressively (10/min to 30/min)
* Agent registry endpoints allowing anonymous access (now require auth)
* Insight engine not supporting Kiro JSONL format (0 sessions analyzed)
* Insight substantive session filter too strict (60s minimum)

## Approach

- **Kiro credits**: Added retry with 300ms delays on stop events for credit file read race
- **MV fix**: Use minIf/maxIf to exclude 2099 sentinel timestamps, backfill existing data
- **Auth**: Detect refresh token in localStorage on new tabs, silently refresh before redirecting
- **Traces UX**: Pure parameterized search bar (platform:, user:, agent:, model:, days:, status:), server-side pagination
- **Insights**: Added Kiro JSONL parsing to session_meta_extractor, relaxed substantive filter
- **Dialog animation**: Removed slide transforms, deferred heavy content render

## How Has This Been Tested?

- Ruff lint passes on all modified Python files
- TypeScript compiles with zero errors
- CLI test suite (18 tests) passes
- Verified ClickHouse data directly to confirm session attribution and MV fixes
- Checked worker logs to trace insight pipeline behavior
image
image

## Learning (optional, can help others)

- ClickHouse AggregatingMergeTree with plain (non-aggregate) columns keeps the first inserted value during merge, not the latest. Use SimpleAggregateFunction for columns that need to update.
- Kiro JSONL uses `{kind, data: {meta: {timestamp}}}` not `{type, timestamp}` like Claude Code.
[...]

PR #1447, opened 2026-06-16
fix(attribution): resolve agent version from lockfile on session push

ShaanNarendran:
## Purpose / Description

Session push hooks were always sending `agent_version=NULL` to the server because `_resolve_agent()` returned early at step 1 (env var) or step 2 (JSONL parsing) before reaching the lockfile lookup in step 3. This caused the Insights page to show "0 sessions available for v1.0.0" even when sessions existed, because ClickHouse had no version tag on any ingested session.

## Fixes

* Fixes version-scoped insights showing 0 sessions despite sessions being attributed to the agent

## Approach

The lockfile (`~/.observal/lockfile.json`) is the authoritative source for which version of an agent is installed. Previously, `_resolve_agent()` had three steps with early returns:

1. `OBSERVAL_AGENT_NAME` env var (Kiro) -> returned `(name, None)` immediately
2. JSONL `agent-setting` line (Claude Code) -> returned `(name, None)` immediately
3. Lockfile lookup -> only reached if steps 1 and 2 both missed

Since Kiro hooks always set the env var, step 3 was never reached for Kiro. Similarly, Claude Code sessions always have the JSONL line, so step 3 was dead code for all first-class IDEs.

The fix decouples name resolution from version resolution:
- Steps 1/2 resolve the agent **name** (no early return)
- The lockfile is **always consulted** for version, regardless of how the name was found
- If the lockfile entry matches the resolved name, the lockfile UUID is also used as the canonical `agent_id`

## How Has This Been Tested?

- 7 new unit tests covering all resolution paths (env var + lockfile, JSONL + lockfile, lockfile-only, no sources, name mismatch, empty cwd)
- Existing cursor session push tests (15) continue to pass
- CLI test suite (32) passes

The regression was introduced in `eb6ab0ae` (PR #1257, May 28) which replaced `read_agent_marker()` with `_resolve_agent()`. The lockfile infrastructure added in `d45ae2e7` (PR #1342, June 3) partially addressed it by adding step 3, but the early returns in steps 1/2 meant no first-class IDE ever reached it.

## Checklist

## AI Assistance

- [x] Yes (Pi): Used to trace the regression through git history and implement the fix

### DIFF
--- a/observal_cli/sessions/base.py
+++ b/observal_cli/sessions/base.py
@@ -447,38 +447,70 @@ def _maybe_upload_layer_snapshot(
 
 
 def _resolve_agent(cwd: str, lines: list[str], session_jsonl: Path | None) -> tuple[str | None, str | None]:
-    """Resolve agent identity from multiple sources (priority order).
+    """Resolve agent identity from multiple sources.
 
-    1. OBSERVAL_AGENT_NAME env var (Kiro per-agent hook commands)
-    2. agent-setting line in JSONL (Claude Code embeds active agent)
-    3. Lock file lookup by cwd + IDE
+    Name resolution (first match wins):
+      1. OBSERVAL_AGENT_NAME env var (Kiro per-agent hook commands)
+      2. agent-setting line in JSONL (Claude Code embeds active agent)
+      3. Lock file lookup by cwd + IDE
+
+    Version resolution always comes from the lockfile. Steps 1 and 2 only
+    provide the agent name; the lockfile is the authoritative source for
+    the installed version.
     """
     import os
+
+    # Resolve agent name from env var or JSONL
+    agent_name: str | None = None
 
     # 1. Env var (Kiro per-agent hooks embed this)
     env_agent = os.environ.get("OBSERVAL_AGENT_NAME", "")
     if env_agent:
-        return env_agent, None
+        agent_name = env_agent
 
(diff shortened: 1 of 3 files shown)


observal_cli/sessions/base.py
```diff
@@ -447,38 +447,79 @@ def _maybe_upload_layer_snapshot(
...
+    try:
+        from observal_cli.lockfile import get_agent_for_directory
+
+        for ide in (
+            "claude-code",
+            "cursor",
+            "kiro",
+            "pi",
+            "codex",
+            "opencode",
```
Haz3-jolt: Bad idea to hardcode. Fetch from ide registry.

tests/test_resolve_agent_version.py
```diff
@@ -0,0 +1,151 @@
...
+        assert agent_version is None
+
+    def test_env_var_name_mismatch_returns_none_version(self, tmp_path):
+        """When env agent name differs from lockfile name, version is None.
+
+        We only attribute a version when the lockfile confirms the agent.
+        """
+        lockfile_entry = {
+            "name": "other-agent",
+            "id": "uuid-other",
```
Haz3-jolt: Use an actual lockfile example. This is not the correct format.

Haz3-jolt (changes requested):

Haz3-jolt (approved):

PR #1455, opened 2026-06-17
feat(sso): add SSO health validators and fix SAML PEM parsing

### LINKED ISSUE #1454: Validation button for SSO setup
### Is your feature request related to a problem? Please describe.

Add a validate button for SSO setup to verify if the url redirect to the respective SSO application functions as intended.

### Describe the solution you'd like

A validate button next to SSO with OIDC and SAML when clicked should return either "SSO functions" message, or the error message concerning the SSO setup. 

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

Lokesh7025:
## Purpose / Description
  Add live SSO health indicators to the login page and admin "Validate" buttons so operators can confirm OIDC/SAML config is working before users hit it. Bundles three fixes uncovered while building the validators: a SAML 500 from malformed PEM parsing, a device-flow OIDC login that ignored the original redirect path, and a false-negative validator that keyword-scanned IdP login pages.
 
  > Note: 100% validation is not possible. Assertion replay, per-user authz, and IdP-only state are not visible server-side, so
   a green result still depends on a real user round-trip.

  ## Fixes
  * Fixes #1454

  ## Approach
  - **OIDC**: probes discovery → authorization endpoint (real redirect_uri/client_id) → token endpoint (deliberately-bad code)
  → JWKS → email scope → issuer/host → token auth methods → clock skew.
  - **SAML**: exercises the real `saml_login` path (decrypt SP key, build OneLogin auth, generate AuthnRequest), then checks
  IdP cert vs metadata, cert expiry, ACS-host vs frontend host, SP cert/key pair, IdP SSO reachability, NameIDFormat.
  - Every probe returns a structured check (`{name,label,status,message,hint}`) so the UI can render per-check pass/fail with
  hints — run-all semantics, no first-fail short-circuit.
  - `/config/sso-health` runs OIDC and SAML concurrently. `Cache-Control: no-store` so refresh re-probes.
  - Core/ee decoupling via a hook registry (`services/saml_health.py`) so the public endpoint exercises enterprise code without
   violating the import boundary.

  ## How Has This Been Tested?
  - Pulled the Okta redirect URI → indicator + validate button went red with the IdP error; re-added → green.
  - Stored a space-separated single-line cert (the broken format) → SAML login no longer 500s.
  - Removed IdP cert → validate showed `idp_cert_or_fingerprint_not_found_and_required`.
  - Verified `Cache-Control: no-store` on `/config/sso-health`.
  - `ruff check`, `ruff format --check`, `pip-audit`, `pnpm audit --prod`, import-boundary grep all clean locally.

  ## Checklist
  - [x] Descriptive commit message
  - [x] Comments only where the *why* is non-obvious
  - [x] Self-review
  - [x] UI screenshots

  ## AI Assistance
  - [x] Yes (Claude Code)
  - [x] Manually reviewed and tested
  
Screenshot 2026-06-18 092819

  
Screenshot 2026-06-18 092916

  
Screenshot 2026-06-18 093009

Screenshot 2026-06-18 092542

Screenshot 2026-06-18 093731


.github/workflows/ci.yml
```diff
@@ -196,6 +196,10 @@ jobs:
 
       - uses: pnpm/action-setup@v6
         if: needs.changes.outputs.web == 'true'
+        with:
+          # Skip the npm-registry install-then-downgrade dance (it hit a
```
Haz3-jolt: Unnecessary comment

ee/observal_server/services/saml.py
```diff
@@ -157,9 +157,52 @@ def build_saml_settings(
...
+    if not metadata_url:
+        return None
+    try:
+        async with httpx.AsyncClient(timeout=10) as client:
+            resp = await client.get(metadata_url)
+            resp.raise_for_status()
+            xml = resp.text
+    except Exception:
+        return None  # metadata not reachable — don't hard-fail an optional check
+    certs = re.findall(r"<[^>]*X509Certificate[^>]*>(.*?)</[^>]*X509Certificate>", xml, re.DOTALL)
```
Haz3-jolt: RegEx is enough for validation?

observal-server/pyproject.toml
```diff
@@ -52,12 +52,13 @@ dependencies = [
 ]
 
 [tool.uv]
-# CVE-2026-48710: starlette <1.0.1 has Host header poisoning.
-# prometheus-fastapi-instrumentator 7.1.0 caps starlette<1.0.0 but is
-# runtime-compatible with 1.x (only a type annotation differs).
+# CVE-2026-48710 / CVE-2026-54282 / CVE-2026-54283: starlette <1.3.1 has
+# Host header poisoning, url-authority confusion, and unbounded urlencoded
```
Haz3-jolt, line 57: Dep audit still fails?

web/src/pages/admin/sso.tsx
```diff
@@ -22,9 +23,113 @@ import { useRoleGuard } from "@/hooks/use-role-guard";
 import { Button } from "@/components/ui/button";
 import { Badge } from "@/components/ui/badge";
 import { Card, CardContent, CardHeader, CardTitle, CardDescription } from "@/components/ui/card";
+import { Tooltip, TooltipContent, TooltipProvider, TooltipTrigger } from "@/components/ui/tooltip";
 import { PageHeader } from "@/components/layouts/page-header";
 import { ErrorState } from "@/components/shared/error-state";
 
+type ValidateResult = {
+  success: boolean;
```
Haz3-jolt: Add optic logs, trace, debug, warning and info level logs.

Haz3-jolt (changes requested):

ee/observal_server/routes/sso_saml.py
```diff
@@ -65,24 +74,34 @@ def _safe_redirect_path(value: str | None) -> str:
...
+        return _env_saml_config_cache
+    async with _env_saml_config_lock:
         if _env_saml_config_cache is not None:
             return _env_saml_config_cache
-
         sp_entity_id = ds.get_sync("saml.sp_entity_id") or f"{_get_frontend_url()}/api/v1/sso/saml/metadata"
         sp_acs_url = ds.get_sync("saml.sp_acs_url") or f"{_get_frontend_url()}/api/v1/sso/saml/acs"
         enc_password = ds.get_sync("saml.sp_key_encryption_password")
         if not enc_password:
-            logger.warning(
```
Haz3-jolt, line 85: Don't add and change in the same PR.

Haz3-jolt (changes requested):

Haz3-jolt:
Not a fan of the UI? If one of the 2 like sso is not working but saml sso is, just show that one. Don't show anything they aren't supposed to use. Similar to insights the red green signal for readiness should be in dynamic settings.

Lokesh7025:
> Not a fan of the UI? If one of the 2 like sso is not working but saml sso is, just show that one. Don't show anything they aren't supposed to use. Similar to insights the red green signal for readiness should be in dynamic settings.

These are validation checks for setup, makes it easier to check if the okta is configured right, so both needs to be shown right? it's not exactly meant for user convenience, it's primarily for the setup, should I keep it only in the sso settings page then? and remove the signal from the login page.

ee/observal_server/routes/sso_saml.py
```diff
@@ -184,27 +340,176 @@
...
+    ``_parse_relay_state``. Re-validate at this boundary as defence-in-depth.
+    Both the path and host are fully server-controlled (``frontend_base`` is
+    deployment config, never request input); only the query value derives
+    from internal state and is encoded via ``urlencode``.
+    """
+    if not sso_diagnostics.is_safe_session_id(corr_id):
+        corr_id = "invalid"
+    base = _get_frontend_url().rstrip("/")
+    qs = urlencode({"sso_error": corr_id})
+    return RedirectResponse(url=f"{base}/login?{qs}", status_code=302)
```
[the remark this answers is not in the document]
Lokesh7025: This is a false positive, no user input can let the url be redirected to another page.

Haz3-jolt (approved):
LGTM

PR #1533, opened 2026-06-29
fix(ee): evaluate enterprise config dynamically on SSO/SCIM requests

shreem26:
## Purpose / Description
The EnterpriseGuardMiddleware evaluated config issues once at boot time using the in-memory sync cache. If `deployment.frontend_url` was set via the UI after the server started, the middleware still blocked SSO/SCIM routes with a stale "localhost or empty" error — requiring an API restart to pick up the change.

## Approach
- Added `validate_enterprise_config_async()` that reads settings from Redis/DB directly (via `ds.get()`), bypassing the stale in-memory `_sync_cache`
- Middleware now calls the async validator on each SSO/SCIM request — these are low-traffic login endpoints so the overhead is negligible
- Middleware is always registered (not conditionally based on boot-time issues)
- Downgraded `saml.sp_key_encryption_password` from a blocking issue to a non-blocking warning since SAML functions without it (the SP key only lives in memory)

## How Has This Been Tested?

- `uv run pytest tests/test_enterprise.py` — all 6 tests pass
- Verified `ds.get("deployment.frontend_url")` returns the correct DB value inside the running container
- Confirmed the API returns 302 (redirect to IdP) on `/api/v1/sso/saml/login` after the fix
- Confirmed the ACS endpoint (`/api/v1/sso/saml/acs`) passes through the middleware without being blocked

## Learning
The dynamic settings system has two tiers:
- `ds.get_sync()` — reads from an in-memory dict, only refreshed on the worker that handles a write
- `ds.get()` — async, reads Redis (30s TTL) → DB → defaults, consistent across all workers

Middleware that gates features should use the async path to avoid stale cross-worker cache issues.

## Checklist

## AI Assistance
Was generative AI tooling used to co-author this PR?

- [x] Yes(Please Specify the tool): Kiro CLI

### DIFF
--- a/ee/__init__.py
+++ b/ee/__init__.py
@@ -23,16 +23,19 @@ logger = logging.getLogger("observal.ee")
 def register_enterprise_middleware(app: FastAPI, settings: Settings) -> list[str]:
     """Register enterprise middleware (must be called before app startup).
 
-    Returns a list of config issues (empty = healthy).
+    Returns a list of config issues at boot (for logging only).
+    The middleware re-evaluates dynamically on each request.
     """
     from ee.observal_server.middleware.enterprise_guard import EnterpriseGuardMiddleware
     from ee.observal_server.services.config_validator import validate_enterprise_config
 
     issues = validate_enterprise_config(settings)
 
+    # Always add the middleware — it evaluates issues dynamically per request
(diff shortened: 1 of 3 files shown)


Haz3-jolt (approved):
LGTM

PR #1551, opened 2026-07-02
fix(cli): relax version enforcement to allow newer CLI with older server

### LINKED ISSUE #1554: CLI version enforcement blocks commands when CLI is newer than self-hosted server
### Checked for duplicates?

1. Self-host an Observal server at version **1.9.5**
2. Install or upgrade the Observal CLI to version **1.9.8**
3. Run `observal auth login --server <your-server-url>` and authenticate successfully
4. Run any CLI command (e.g. `observal agent list`, `observal scan`, etc.)
5. The CLI rejects the command, stating the server version is 1.9.8 (which it is not — the server is 1.9.5)

The CLI appears to cache a `server_version` in `~/.observal/version_cache.json` with `"source": "server"`. The server exposes a `misc.max_cli_version` dynamic setting via its config endpoint. When the CLI version (1.9.8) exceeds the self-hosted server version (1.9.5), the version enforcement kicks the user out.

### Expected behaviour

- The CLI should ping the **user's configured server**, retrieve its actual running version, and use that for compatibility checks — not a cached or centrally-advertised version.
- A CLI newer than the server should **warn** but not **block** commands, unless there is a known breaking API change between those specific versions.
- Every new CLI release should not break all self-hosted users who have not yet upgraded their server image.
- If there is a genuine incompatibility, the error message should clearly state: "Your server is running version X, but this CLI requires at least version Y. Please upgrade your server."

### Support bundle

_No support bundle attached yet._

**Evidence from investigation:**

- `~/.observal/version_cache.json` contains:
  ```json
  {
    "latest_version": "1.9.8",
    "source": "server",
    "server_version": "1.9.8"
  }
  ```
- Server logs show repeated reads of `misc.max_cli_version` dynamic setting on every API call (`api.routes.config:get_version` → `services.dynamic_settings:get` → `misc.max_cli_version`)
- The `update_check_repo` config key is empty, confirming the check is not going to a central/public registry

**Suggested fix:**

The version compatibility logic should:
1. Query the connected server's actual version on login (or periodically), not assume CLI version == required server version
2. Compare CLI version against the server's declared `min_cli_version` / `max_cli_version` range
3. Only hard-block if the server explicitly declares the CLI version incompatible
4. Default to a warning for minor version mismatches (e.g. CLI 1.9.8 vs server 1.9.5)

VishnuM049:
## Purpose / Description
The version enforcement gate required an exact match between CLI and server versions. This broke self-hosted deployments every time a new CLI release shipped, because users get the latest CLI (e.g. 1.9.8) but their server stays pinned at deployment version (e.g. 1.9.5). The CLI refused to connect with "CLI version X is ahead of server Y."

The problem was not where versions are fetched (it correctly pings the target server), but the policy: exact equality is too strict when servers are independently deployed.

## Fixes
* Fixes #1554

## Approach
Changed the version compatibility policy from "exact match" to "forward-compatible":

- Same major, CLI >= server: allowed. A newer CLI can talk to an older server since we maintain backward compatibility within a major version.
- CLI behind server (same major): blocked. The server may expose APIs the CLI does not understand.
- Different major version: blocked. Major bumps signal breaking API changes.

Three enforcement points updated:
1. `observal_cli/cmd_auth.py` (`_ensure_cli_matches_server`): the login gate
2. `observal_cli/version_check.py` (`check_version_compatibility`): the per-session gate used by all commands via `client.py`
3. `observal-server/middleware.py` (`version_middleware`): the server-side HTTP middleware that returns 426 on incompatible requests

Also fixed a pre-existing test bug where tests imported `click.exceptions.Exit` but typer raises a RuntimeError subclass.

## How Has This Been Tested?

- `python3 -m pytest tests/test_version_check.py` (52/52 pass)
- Server-side middleware tests have correct logic but require Docker deps (`slowapi`) not available locally; syntax verified with `ast.parse()`
- Verified CLI-ahead scenario (1.9.8 vs 1.9.5) now passes, CLI-behind and major mismatch still block

There are two enforcement layers:
1. CLI-side: checks before login and once per session. This is what users see when blocked.
2. Server-side middleware: checks every authenticated HTTP request from the CLI via the `X-Observal-CLI-Version` header. Returns HTTP 426 if incompatible. Auth, ingest, and telemetry paths are exempt. Browser requests skip this entirely.

Both had the same exact-match bug. Both are now relaxed to the same forward-compatible policy.

## Checklist

## AI Assistance

- [x] Yes(Please Specify the tool): Kiro (Claude Opus 4.6)

### DIFF
--- a/observal-server/middleware.py
+++ b/observal-server/middleware.py
@@ -228,7 +228,11 @@ def configure_version_middleware(app: FastAPI) -> None:
                 try:
                     client_ver = Version(cli_ver_str)
                     server_version = Version(server_ver)
-                    if client_ver != server_version:
+                    # Allow newer CLI within same major (forward-compatible).
+                    # Block if CLI is behind or major version differs.
+                    if client_ver.major != server_version.major:
+                        return _cli_version_response(server_ver, cli_ver_str)
+                    if client_ver < server_version:
                         return _cli_version_response(server_ver, cli_ver_str)
                     effective = str(server_version)
                 except InvalidVersion:
--- a/observal-server/tests/test_config.py
+++ b/observal-server/tests/test_config.py
@@ -51,7 +51,7 @@ def test_demo_env_vars_default_to_none():
 
 
 def test_version_middleware_rejects_cli_drift(monkeypatch):
-    """CLI requests must match the exact server version."""
+    """CLI requests behind server version should be rejected."""
     import version
     from middleware import configure_version_middleware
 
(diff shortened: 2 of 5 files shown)


Haz3-jolt (approved):
LGTM

HemalathaMadeswaran18:
@VishnuM049 @Haz3-jolt the point was that the enforcement should exist between the server its connected to and the CLI. not the latest image version. I dont think Same major, CLI >= server: allowed. addresses that. please revisit

PR #1588, opened 2026-07-10
fix: correctly update browser tab favicon when logo is changed dynamically

ShaanNarendran:
## Purpose
The favicon (browser tab icon) was not updating immediately when a super admin changed the branding logo dynamically in the settings.

## Approach
The previous implementation grabbed only the very first  and merely replaced the . This fails for two reasons:
1. It misses  and .
2. Chrome and Safari heavily cache favicons and usually ignore changes on an existing node.

The fix resolves this by querying all  links in the head, and explicitly tearing down the old nodes while injecting freshly created ones. This accurately forces an immediate repaint in all major browsers.

## How Tested
Tested locally in Chrome and Safari by visiting Deployment Settings and dynamically changing the App Logo URL. Verified the tab icon updates immediately without requiring a hard refresh.

## Checklist
- [x] Tested locally
- [x] Code is clean and conforms to project standards
- [x] Readme / Documentation updated if needed (N/A)


web/pnpm-lock.yaml
```diff
@@ -0,0 +1,5887 @@
+lockfileVersion: '9.0'
+
```
Haz3-jolt: Uncommit

Haz3-jolt (changes requested):

Haz3-jolt (approved):

PR #1624, opened 2026-07-23
Feat/registry namespace display

Lokesh7025:
## Purpose / Description

1. Registry identities rendered as raw `namespace/slug` everywhere, so listings read as `lokeshselvam7025/task-creator`. Hard to scan.
2. A username that isn't a valid namespace could never publish (`namespace_for_user()` rejects it) *and* could never be changed (`user_has_listings()` counts soft-deleted agents). Permanent deadlock, with no way out short of deleting the account.

## Approach

**Display** bare name on the first line, owning namespace beneath it as `@handle` in a smaller muted font.

- Web: `registryIdentity()` + a shared `<RegistryName>`, used by agent/component cards, both index tables, both detail headers, the leaderboard, and home dashboard rows.
- CLI: `name_block()` for table cells (two lines), `name_inline()` for panel titles and `--output plain`.
- Copy-pasteable commands are untouched  install/pull commands, the reference resolver, and `client.canonical_name()` still emit canonical `namespace/slug`.

**Dots**  permitted in namespaces, so handles predating namespace validation stay usable. No migration needed: the charset was only ever enforced in Python. There are no CHECK constraints on any identity table and the columns are plain `varchar(32)` 
Postgres already accepted dots, which is how these handles got in.

- The rule lived in four places that had to agree; now one Python home in `observal_shared.namespace_rules`, imported by both the server validator and the CLI pre-check. Web mirrors it in `registry-name.ts`.
- Consecutive dots stay rejected  a namespace reaches local install names and harness config keys.
- Dots are flattened out of local install names, matching how `lockfile.local_registry_name` already flattens the registry hostname. Flattening lets `a.b` and `a-b` collapse onto one qualifier, so the server-side builder now keeps collided names distinct.

**Publish lock** unchanged for valid usernames (`test_username_change_is_blocked_after_publish` passes untouched). It's lifted only for usernames that could never publish; their listings move namespace with the rename. Namespace mutation isn't new: `transfer_entity_owner()` already reassigns namespaces and `lockfile_reconcile` already syncs clients.

## How Has This Been Tested?

Against a local Docker stack (`make rebuild`, API + web + db + clickhouse + redis healthy).

- CLI: 64 passed. Verified every list surface (agents + all five component types), detail panels, `--output plain`, and that `registry mcp show super/obsv-test-echo` still resolves by the canonical slash form.
- Server: 348 passed. 26 pre-existing failures (`KeyManager not initialized`) confirmed identical on `origin/main` in a scratch worktree.
- Dot handling verified in the running API container: dotted handle accepted, case normalised, `..` / leading dot / trailing dot / space rejected, hyphen handles unaffected.
- `pnpm typecheck`, `pnpm lint`, `ruff check`, `ruff format --check` all clean.

## Checklist

## AI Assistance

- [x] Yes (Please Specify the tool): Claude Code

image

image

### DIFF
--- a/observal-server/api/routes/auth.py
+++ b/observal-server/api/routes/auth.py
@@ -52,7 +52,7 @@ from schemas.sso_health import make_check
 from services import sso_diagnostics
 from services.jwt_service import create_access_token, create_refresh_token, decode_access_token, decode_refresh_token
 from services.redis import get_redis
-from services.registry_namespace import user_has_listings
+from services.registry_namespace import is_valid_namespace, rename_namespace, user_has_listings
 from services.security_events import (
     EventType,
     SecurityEvent,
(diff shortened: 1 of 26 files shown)


Haz3-jolt (approved):
LGTM

PR #1664, opened 2026-08-07
fix(build): pre-create ~/.observal before Docker makes it root-owned

RAWx18:
## Purpose / Description

On Linux, a first-time `make up` / `make rebuild-fast` breaks the CLI before it is ever used. The compose stack bind-mounts `~/.observal/logs` into the API container, and when that path does not exist yet, the root Docker daemon auto-creates `~/.observal` owned by `root:root`. The first `observal auth login` then crashes with an unhandled `PermissionError` traceback while saving `~/.observal/config.json` — a confusing first-run experience for every new Linux contributor (macOS is unaffected because Docker Desktop creates bind-mount sources as the host user).

A secondary effect of the same root cause: the containerized API (running as `appuser`, uid 1001) could never write `dev.log` into the bind mount on Linux, so `observal logs` silently streamed nothing.

## Fixes
* Fixes #

## Approach

Two small, non-destructive changes:

1. **Makefile** — new `ensure-host-dirs` target that runs before every target that starts the stack (`up*`, `rebuild*`, `reset*`). It pre-creates `~/.observal/logs` as the invoking user (so the Docker daemon never gets the chance to create it root-owned) and grants the container's `appuser` (uid 1001, see `docker/Dockerfile.api`) write access via a narrow `setfacl -m u:1001:rwx` — no world-writable permissions. The ACL step is skipped silently where `setfacl` is unavailable (macOS doesn't need it: Docker Desktop maps container uids to the host user) or unsupported (falls back to today's behavior).
2. **observal_cli/config.py** — `config.save()` now catches `PermissionError` and exits with a clear, actionable message (including the exact quoted `sudo chown -R "$USER" ~/.observal` command) instead of dumping a 60-line traceback. `SystemExit` matches the CLI's existing fatal-error convention in `client.py`.

Existing installs are unaffected: the recipe is idempotent, cannot fail the build (`|| true`), and production deployments (`install-server.sh` / server package) have no `~/.observal` bind mount at all.

## How Has This Been Tested?

Reproduced and verified end-to-end on Fedora Linux (rootful Docker, host uid 1000):

- **Repro:** fresh machine, `make rebuild-fast` → `~/.observal` created `root:root` → `observal auth login` crashes with `PermissionError`. After this fix (dir pre-created user-owned), login succeeds and the config is saved.
- **ACL verified:** `getfacl ~/.observal/logs` shows `user:1001:rwx` with `other::r-x` (no world-write); confirmed the running API container can append to `dev.log` through the bind mount (`docker exec … sh -c 'echo ok >> …/dev.log'`).
- **Makefile:** `make -n up` / `make -n rebuild-fast` show correct ordering (mkdir + setfacl before `compose up`); `make help` output unchanged; recipe is idempotent across repeated runs.
- **Test suites:** `make test` → 5346 passed, 20 skipped; `observal_cli/tests/` → 78 passed. The 23–24 failures in `observal-server/tests/` are pre-existing on `main` (verified by stashing this change and re-running).
- **Lint:** `ruff check` and `ruff format --check` clean.
- **Error path:** no caller catches `PermissionError` around `config.save()` (verified by grep), and all tests mock it, so the new `SystemExit` changes no tested behavior.

- On Linux with rootful Docker, the daemon creates missing bind-mount *source* directories as `root:root` — there is no compose option to change this, so pre-creating the path on the host is the standard fix.
- `setfacl -m u:<uid>:rwx` is a tighter alternative to the usual `chmod 1777` for sharing a bind mount with a non-root container user: it grants exactly one uid and avoids a world-writable directory in `$HOME`. (The classic sticky-dir attacks are also kernel-mitigated nowadays via `fs.protected_symlinks/regular`, but not needing the mitigation is better.)
- Handy recovery without sudo for already-broken dirs: `docker run --rm -v "$HOME/.observal:/fix" alpine chown -R "$(id -u):$(id -g)" /fix`.

## Checklist

- [ ] UI changes: include screenshots of all affected screens (in particular showing any new or changed strings) — N/A, no UI changes

## AI Assistance

- [x] Yes(Please Specify the tool): GitHub Copilot (agent mode, Claude)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved startup, rebuild, and reset workflows by preparing required host log directories and access permissions automatically.
  * Added clearer guidance when configuration files cannot be created, read, or updated because of permission issues.
  * Preserved existing configuration merging and secure file-permission behavior.

* **Security**
  * Updated cryptographic protection to address a vulnerability affecting PKCS#7 decryption.
  * Updated YAML processing and diagram rendering for improved compatibility and reliability.


Haz3-jolt (approved):
LGTM

VishnuM049 (approved):

VishnuM049 (approved):

PR #1669, opened 2026-08-07
feat(inbox): add the cross-product actionable inbox

Lokesh7025:
## Purpose / Description

There is no inbox or notification system today. A reviewer has no durable signal that something is waiting on them, a submitter learns a decision only by re-opening the listing, and `observal outdated` findings disappear the moment the terminal scrolls.

This adds a per-user work feed covering review assignments, decisions on your own submissions, and installed-item update notices, readable from the web and the CLI, with action history that survives being read.

## Approach

Two Postgres tables, `inbox_items` and `inbox_item_events`, with one row per recipient per event. Fan-out happens on write so per-user state is first class.

Four decisions carry most of the design:

**Delivery is transactional, not evented.** `EventBus.emit()` awaits handlers inline and swallows their exceptions, logging that the handler's side-effects are lost. That is fine for audit, where a gap is tolerable, but a lost handler here would be a lost inbox item. `services/inbox/deliver()` writes rows in the originating request's transaction instead, so a notice commits exactly when the thing it describes commits.

**Idempotency is a unique constraint, recovered inside a SAVEPOINT.** The key is `(user_id, dedupe_key)`. The savepoint matters: because delivery runs in the caller's transaction, catching `IntegrityError` with a plain `db.rollback()` would discard the caller's own uncommitted work, so a redelivered approval notice would undo the approval, along with every sibling recipient already flushed. Postgres also refuses further statements after an error unless a savepoint absorbs it. There is a regression test that fails against the plain-rollback version.

**Read is not resolved.** `read_at` and `state` are independent columns. An item that has been read is still open work, still counted by the needs-action filter, and still in the default list.

**Authorization is reused, not re-derived.** `recipients.reviewers_for` is the set inverse of `teamspace.can_review`, and a test asserts the two agree across a matrix of callers so they cannot drift. Since membership can be revoked after an item is delivered, items carry `team_id` and `is_private_subject`, and the read paths re-check team membership, matching `check_listing_visibility_async` rather than the review predicate (a plain member may see a private listing they cannot review). Items that no longer resolve are omitted rather than redacted, because a placeholder still leaks that something exists.

Sources wired in this PR:

| Kind | Fired from | Goes to |
| --- | --- | --- |
| `review_requested` | 5 component submits, 5 publishes, agent version release, component version release, both visibility-flip re-queues, teamspace ownership transfer | Everyone who can clear it |
| `review_approved` / `review_rejected` | All 7 decision endpoints in `api/routes/review.py` | The submitter |
| `update_available` | `observal outdated`, reported to `POST /api/v1/inbox/outdated-report` | The reporting user |

Auto-approved publishes emit nothing, since there is no pending state for a reviewer to act on. Agent decisions address the version's `released_by`, which is not always the agent's creator.

Update notices stay CLI-detected because the server cannot compute them alone: `AgentDownloadRecord` knows who installed an agent but not which version, and `ComponentDownloadRecord` knows the version but not the user. The lock file is the only source with both. A newer latest version is a new dedupe key, and the previous notice is closed as superseded rather than deleted.

Kinds whose source does not exist yet (teamspace join requests, review conversations) are declared in the enum but unused, so the enum and its migration stay stable when those land.

Retention purges only resolved items after `inbox.retention_days` (default 90). Open items are never purged on age, because deleting unactioned work silently is the failure this feature exists to prevent.

## How Has This Been Tested?

**Unit,** `tests/test_inbox.py`, 22 tests against a real in-memory SQLite database rather than a mocked session, since the behaviour under test is the database's (unique constraint, `IntegrityError`, savepoint):

```
cd observal-server
uv run --with pytest --with pytest-asyncio --with aiosqlite pytest ../tests/test_inbox.py -q
```

Covers idempotency, batch isolation on collision, the savepoint regression, supersede, read-not-resolving, history append-only, recipient agreement with `can_review`, read-time membership revocation, and the retention predicate.

**Live API,** every endpoint exercised over real HTTP against a running server, not just through tests. 48 of 48 checks passed: fan-out to reviewers and admins, submitter excluded from their own submission, re-publish not duplicating, read leaving the item in the needs-action list, done/dismiss/reopen with history, cross-user access answering 404, supersede on a newer version, filters, filter-scoped read-all, and 401 when unauthenticated.

**CLI,** `observal inbox`, `inbox count`, `inbox --action-required --output json`, and an invalid `--kind` all driven against that server. JSON output parses, tables render, bad input gives a clean error.

**Browser,** logged in as the demo reviewer and confirmed marking an item read moved unread from 2 to 1 while needs-action stayed at 1 and the item stayed in the list, with a `read` entry appended to history.

**Migration,** upgrade, downgrade, and re-upgrade on Postgres. Downgrade drops both tables and both enum types cleanly.

**Lint and typecheck,** repo-wide `ruff check .` and `ruff format --check .`, `pnpm run lint`, `pnpm run typecheck`, and `reuse lint` all clean.

**Regression check,** ran the full suite against a clean baseline worktree at the parent commit and diffed the failure sets. This branch introduced 15 failures, all fixed. The remaining local failures are identical on the baseline and are Windows-only path, symlink, and `fcntl` issues that do not occur on the Linux CI runner.

Several existing tests hand-scripted a fixed number of `db.execute()` results or asserted exact flush and commit counts. Delivery adds recipient queries and a savepoint flush to those paths, so those harnesses now fall back to empty results past their script and the count assertions became lower bounds.

The savepoint point generalises: any helper that writes inside a caller's transaction cannot recover from `IntegrityError` with a session-level rollback. `services/user_recommendations.py::record_feedback` does exactly that and is correct only because it owns its transaction and commits at top level. Copying that shape into a shared-transaction helper is a silent data-loss bug, which is why the pattern here is `begin_nested()` and why there is a test that fails without it.

- [US Copyright Office, Part 2: Copyrightability](https://www.copyright.gov/ai/Copyright-and-Artificial-Intelligence-Part-2-Copyrightability-Report.pdf) informed how the AI assistance below is declared
- SQLAlchemy nested transactions: https://docs.sqlalchemy.org/en/20/orm/session_transaction.html#using-savepoint

## Checklist

## AI Assistance

- [x] Yes (Please Specify the tool): Claude Fable 5, via Claude Code

## New Features

- Added a personal Inbox for review requests, publishing updates, review decisions, system notices, insight reports, and outdated component alerts.
- View unread, actionable, open, completed, and dismissed items with filters, details, history, links, and commands.
- Mark items read/unread, complete, dismiss, reopen, or bulk-mark as read.
- Added Inbox navigation with unread counts and the `observal inbox` CLI.
- Added optional outdated-item reporting and automatic retention cleanup.
- Review tabs can now be opened directly through URL parameters.


Haz3-jolt (approved):

HemalathaMadeswaran18:
LGTM

HemalathaMadeswaran18 (approved):

PR #1706, opened 2026-08-29
Replace ClickHouse with embedded DuckDB

SrihariLegend:
## Purpose / Description

Replace ClickHouse with **embedded DuckDB** as the analytics store (session telemetry, audit/security logs, webhook deliveries, layer snapshots, session rollups). Motivation: remove the ClickHouse dependency (Yandex lineage), shrink the deployment profile (one less container, no 2GB RAM budget, no extra image), and improve query latency for the dashboard/trace workloads.

DuckDB is embedded in the API process: no server, no HTTP hop, no credentials. Backup becomes "copy one file".

## Fixes
* N/A — no tracking issue (dependency replacement / supply-chain decision)

## Approach

- **`observal-server/services/duckdb/`** — the real implementation:
  - `client.py`: process-wide connection (RLock + `asyncio.to_thread`), `_query()` returning an httpx-shaped `QueryResult`, and a **compatibility layer** that auto-translates legacy ClickHouse syntax (`{name:Type}` placeholders, `Array(...)` params, `FINAL`, `FORMAT JSON`, trailing `SETTINGS`, `count()`, `now64(3)`, `toUInt64`, `INTERVAL {n:Type} UNIT`). Unused named params are dropped (DuckDB rejects them; legacy callers pass them).
  - `insert.py`: batch inserts via Arrow relations — DuckDB's `executemany` and row-at-a-time `INSERT OR REPLACE` against ART primary keys are pathologically slow (~200–2k rows/s measured); the Arrow scan + set merge path sustains **~30k rows/s**. `ReplacingMergeTree` semantics preserved via PRIMARY KEY + `INSERT OR REPLACE` (write-time dedup — no `FINAL` anywhere).
  - `migrations.py`, `schema.py`: versioned SQL runner (`duckdb/migrations/001_baseline.sql` consolidates legacy CH migrations 001–004), runtime init, resource pragmas (`memory_limit`, `threads`).
- **`services/clickhouse/`** — deprecated shim re-exporting the DuckDB implementation under legacy names; all ~31 callers work unchanged.
- **Retention**: DuckDB has no table TTL, so the retention cron now explicitly applies the legacy DDL TTLs (730d audit/security logs, 30d `raw_line` scrub).
- **Topology**: DuckDB is single-process read-write → `API_WORKERS=1` and the arq worker runs **embedded in the API process** (`EMBEDDED_WORKER=true`, default); the separate worker service and the ClickHouse service are removed from compose.
- **Migration jobs**: ClickHouse export/import scopes are rejected with a clear error (DuckDB archive export/import is follow-up work).
- **Deferred (documented in `docs/duckdb-analytics.md`)**: observal_cli installer, server-package compose, Grafana dashboards.

## How Has This Been Tested?

- **Unit/integration**: new live-engine suites `tests/test_duckdb_analytics.py` (19 tests: migrations idempotency, write-time dedup, checkpoint rewind, rollup correctness, retention TTLs, concurrency, compat layer) and `tests/test_duckdb_resource_tuning.py` (11 tests, live pragma verification). Root suite: **8,235 passed, 0 failed**; server-suite failure set byte-identical to `main` (12 pre-existing failures).
- **Live E2E**: booted Postgres 18 + Redis 8 + API; login → session ingest (5 JSONL lines) → replay/summary/dashboard endpoints → re-ingest dedup (`skipped: 5`, still 1 session) → live audit trail → embedded arq cron firing. One real integration bug found and fixed (`arq.create_worker` import + signal handling).
- **Full docker compose stack** of this branch ran healthy for hours with live WAL activity; Playwright E2E suite run against it — all failures traced to pre-existing spec drift (missing `/telemetry/hooks` endpoint, missing fixture users, nonexistent DELETE routes), **zero failures attributable to the port**.
- **Benchmark** (`scripts/bench_duckdb_vs_clickhouse.py`, reproducible): 1M session events + 500k audit rows, vs a real ClickHouse 26.6 container:

| Metric | ClickHouse | DuckDB | Δ |
| --- | --- | --- | --- |
| Ingest | 19.9k rows/s | 30.2k rows/s | **1.5× faster** |
| Session replay p50 | 5.2ms | 3.0ms | **1.7× faster** |
| Dedup range scan p50 | 5.1ms | 2.4ms | **2.1× faster** |
| Dashboard rollups p50 | 9.6–10.8ms | 1.3–1.8ms | **7–8× faster** |
| Recent activity p50 | 9.7ms | 0.9ms | **10× faster** |
| Audit 7d scan p50 | 4.8ms | 3.5ms | **1.4× faster** |
| Storage (1M events + 500k audit) | 110MB | 347MB | CH 3.2× smaller (ZSTD vs FSST) |

To reproduce: `docker run -p 18123:8123 -e CLICKHOUSE_PASSWORD=clickhouse clickhouse/clickhouse-server:26.6`, then from `observal-server/`: `DUCKDB_PATH=/tmp/bench.duckdb python ../scripts/bench_duckdb_vs_clickhouse.py --sessions 2000 --events 500 --audit-rows 500000`.

**Screenshots** (live stack on this branch, real ingested session — a Pi coding session with 1.2M input tokens / 361 tool calls):

Executive dashboard off DuckDB rollups:
dashboard

Traces list with populated token/tool columns:
traces

Trace detail — session replay with dedup ("953 events, deduped from 954"):
trace detail

Audit log writing to and reading from DuckDB:
audit log

- DuckDB's Python `executemany` is a per-row prepared-statement loop (~2k rows/s) and row-at-a-time `INSERT OR REPLACE` on ART-indexed tables is ~200 rows/s; staging via `pyarrow` relations + one set-merge is the fast path (30k+ rows/s).
- ClickHouse `FINAL` (merge-on-read for `ReplacingMergeTree`) is replaced by PRIMARY KEY + write-time dedup; readers get exact ART point-lookups instead of bloom-filter-assisted granule scans.
- DuckDB FSST string compression is ~3–4× on repetitive JSON vs ClickHouse ZSTD(3) at ~18× — the one regression; acceptable at this workload's scale.
- DuckDB rejects unused named parameters and binds `Array` params differently than ClickHouse HTTP; the compat layer handles both.
- DuckDB is single-process read-write, which is why the arq worker moved in-process (`EMBEDDED_WORKER`).

## Checklist

## Licenses

| Library | Description | License |
| --- | --- | --- |
| duckdb | Embedded analytical database (Python client) | [MIT License](https://opensource.org/licenses/MIT) |
| pytz | Required by duckdb for TIMESTAMPTZ→datetime conversion | [MIT License](https://opensource.org/licenses/MIT) |

## AI Assistance

- [x] Yes(Please Specify the tool): Pi coding agent — drafted the port, tests, and benchmark under human direction

## Discord username (optional)

Discord username:

## Summary by CodeRabbit

* **New Features**
  * Replaced the external analytics database with an embedded DuckDB store.
  * Added configurable database paths, read-only mode, migrations, retention, and resource tuning.
  * Embedded background workers into the API service for simpler deployment.
  * Added support-bundle reporting and benchmarking for the new analytics store.

* **Bug Fixes**
  * Improved handling of empty or null analytics results.
  * Prevented duplicate baseline snapshots through replacement semantics.
  * Improved audit-log response parsing and telemetry retention handling.

* **Documentation**
  * Added deployment, architecture, compatibility, and benchmarking guidance for DuckDB analytics.


amogh-dongre:
Hey @SrihariLegend! 
Nice work on the Arrow insert path and storage changes. A few things I'd like to add/ask from you:

1. **Backend selection:** DuckDB's single-process RW constraint means we can't scale the API the same way as before, while the Helm setup still assumes the existing ClickHouse/worker topology. Can we make this selectable via something like `ANALYTICS_BACKEND=clickhouse|duckdb` instead of making DuckDB a hard replacement?

2. **Migration/rollback:** I don't see a ClickHouse → DuckDB migration or a clear rollback path yet. Since archive export/import is still TODO,you mentioned env flags but that adds needless complexity IMO better option would be to run duckDB and Clickhouse as separate builds in AnalyticsBackend=DuckDB/Clickhouse.

3. **Global lock:** The process-wide `RLock` looks like it will serialize DB access, so a slow query could block unrelated requests. Is it possible to narrow the lock scope or add a concurrent-load benchmark to validate this?

4. **Write semantics:** `INSERT OR REPLACE` doesn't seem equivalent to `ReplacingMergeTree` if we're relying on a version column. Can we add a test for out-of-order writes / duplicate PKs and verify the result matches ClickHouse?

5. **NULL/defaults:** Can we add a test for missing fields? It looks like they may become `NULL` rather than using the column `DEFAULT`, which could change existing behavior.

6. **Audit timestamp:** The `or` condition for the audit timestamp looks inverted. A small test for both the supplied and missing timestamp cases should catch this.

7. **Benchmark:** The current benchmark is serial p50. Given the new global lock, can we also get concurrent throughput + p95/p99 numbers?

8. **Unrelated changes:** Could we split the test/agent fixes into a separate PR? It'll make this change much easier to review.

PR #1722, opened 2026-09-10
Fix/pi extension migrations

### LINKED ISSUE #1602: Auto-install the Pi telemetry extension locally and report stale versions during authentication
## Problem

Pi telemetry currently depends on `npm:observal-pi` being configured in `~/.pi/agent/settings.json`.

Users who do not install the npm package do not receive the telemetry extension automatically. Doctor reports the missing package, but authentication does not provide a local extension installation path.

There is also no authentication-time check that reports when an installed npm or locally managed extension is older than the current Observal CLI release.

## Proposed change

Bundle the canonical Pi extension source with the Observal CLI distribution.

During successful authentication, when Pi is detected:

1. Check whether `npm:observal-pi` is configured.
2. If the npm package is configured, do not create a local duplicate.
3. If the npm package is not configured, atomically install the bundled extension at `~/.pi/agent/extensions/observal.ts`.
4. Record a machine-readable Observal extension version in the managed file.
5. Compare the active npm or managed local extension version with the current CLI version.
6. If the npm extension is older, keep login successful and print an actionable `pi update npm:observal-pi` reminder.
7. If an Observal-managed local copy is older, refresh it atomically from the current CLI bundle and tell the user to reload Pi.
8. If the target file exists but is not marked as Observal-managed, do not overwrite it. Report the conflict with manual resolution instructions.

The repository should retain one canonical extension source rather than maintaining separate TypeScript copies.

## Related integration changes

- Include the extension source in the CLI wheel.
- Update authentication setup to install or check the extension.
- Update doctor to recognize npm and local installations.
- Update the Pi adapter's hook detection to recognize the local managed extension.
- Surface copy, parsing, and version-check failures instead of swallowing them.
- Document both installation modes.

## Acceptance criteria

- [ ] Authentication injects the extension when Pi exists and no npm package is configured.
- [ ] Parent directories are created as needed.
- [ ] The file is written atomically.
- [ ] An npm installation prevents duplicate local injection.
- [ ] A configured but not yet downloaded npm package still prevents duplicate injection.
- [ ] The managed local file contains a parseable extension version.
- [ ] Older npm installations produce an actionable update reminder.
- [ ] Older managed local installations are refreshed and report that Pi must reload.
- [ ] Current versions produce no warning.
- [ ] Newer versions are not downgraded.
- [ ] Unmanaged local files are not overwritten.
- [ ] Doctor recognizes npm and local installation modes.
- [ ] Pi adapter hook detection recognizes npm and local installation modes.
- [ ] Authentication remains successful when only a stale-version reminder is needed.
- [ ] Tests cover missing, current, stale, newer, npm-configured, and unmanaged-file cases.
- [ ] A built wheel contains the Pi extension source.

amogh-dongre:
## Purpose / Description

The Pi telemetry extension install path added for #1602 shipped two defects that
affect users upgrading from earlier CLI versions.

**1. Pre-tracking installs were disowned instead of upgraded.** CLI versions before
the install manifest wrote `~/.pi/agent/extensions/observal.ts` with nothing beside
it. Once the bundled extension source changed, those files matched neither the
manifest check nor the exact-content check, so `doctor` reported them as unmanaged
and asked the user to delete them by hand. The behaviour before #1602 refreshed them
silently. Every existing local install would have hit this on its next upgrade.

**2. A local extension alongside a configured npm package sent every session twice.**
`check_status` returned as soon as it found `npm:observal-pi` in `settings.json`,
before it ever looked at `extensions/observal.ts`. Pi loads both channels, so a file
left over from local mode kept pushing sessions in parallel with the npm package —
and `check`, `patch` and `cleanup` were all blind to it. Earlier releases prevented
this by stripping the npm registration; that removal was dropped when npm became the
winning mode, with nothing put in its place. `packages/pi-extension/README.md` still
promised the old behaviour, so the documentation had quietly become false.

## Fixes

* Fixes #1602

## Approach

Both defects come down to the same missing capability: recognising a local
`observal.ts` as one Observal wrote when no manifest says so.

Every `observal.ts` this project has shipped carries the same doc header — verified
against all 19 revisions in the file's history. `_is_observal_authored()` matches on
it, and a regression test asserts the current bundled source still contains it, so
rewording that header fails loudly rather than silently disabling recognition.

With that in place:

- **Migration** — an untracked file carrying the header is refreshed in place and the
  previous copy kept as `observal.ts.bak`. A file *without* the header is still never
  written to.
- **De-duplication** — in npm mode, a local copy Observal recognises (by manifest or
  header) is reported by `doctor check` and removed by `doctor patch` and
  `doctor cleanup`, keeping a `.bak`. The npm registration in `settings.json` is never
  touched, and neither is a file Observal did not write.
- **Backups** — the header also matches a copy someone has edited, so both paths copy
  the file aside before overwriting or deleting. The backup name does not end in `.ts`
  so Pi will not load it, and collisions fall back to `.bak.1`, `.bak.2`.
- **Login stays non-destructive** — `auth login` performs migrations but only *warns*
  about a duplicate, since deleting a file mid-login is too surprising. It points at
  `observal doctor patch --harness pi`.

This also resolves a symptom of the issue's "surface parsing failures" item: a corrupt
manifest or unparseable version beside a recognised file now migrates rather than
reporting the file as unmanaged.

`packages/pi-extension/README.md` and `docs/reference/config-files.md` are corrected
to describe both installation modes as they actually behave.

## How Has This Been Tested?

Environment: Python 3.11, `.venv`, `pytest`. Server-side suites are unaffected — this
touches CLI modules only.

**Blast radius** — every test file referencing `pi_extension`, the doctor Pi
functions, the auth Pi setup, or the harness adapters:

**Coverage** — `observal_cli/pi_extension.py` at 95% (187 statements, 10 missed). All
ten misses are defensive `except OSError` paths.

**Mutation testing** — twelve behaviours were broken one at a time to confirm the new
tests are load-bearing rather than padding. Ten of twelve were caught, e.g. disabling
the header check kills 20 tests; making it match everything kills 11; removing
duplicate detection kills 9; skipping either backup kills 3 and 2.

Two mutations survived and are known gaps: the manifest branch of
`_is_observal_local_install` is never the deciding factor in any test, and the
manifest-removal assertion in `test_dedupe_removes_the_local_copy_and_keeps_a_backup`
passes vacuously because its fixture creates no manifest.

**Added tests** — 30 new test functions: 20 in `test_pi_extension.py`, 8 in
`test_cmd_doctor.py`, 2 in `test_cmd_auth.py`. They cover migration, de-duplication,
dry runs, backup collision handling, idempotency, `settings.json` preservation,
cleanup, stale-npm-plus-duplicate, and — importantly — that a foreign file is left
untouched by every path.

**Manual verification** — simulated a pre-tracking user upgrading and an npm user with
a local leftover, confirming in both cases that the file is refreshed or removed, the
backup matches the original byte for byte, `settings.json` is unchanged, only
`observal.ts` remains loadable by Pi, and a second run is a no-op.

The two defects share a root cause worth recording: when an installer moves from
"overwrite unconditionally" to "only touch what we track", every pre-existing install
becomes untracked overnight. Adding the tracking metadata is the easy half; deciding
what to do with installs that predate it is the half that actually reaches users.

Mutation testing was what surfaced the vacuous assertion above — ordinary coverage
reported those lines as covered.

## Checklist

- [ ] UI changes: include screenshots of all affected screens — n/a, CLI only

## AI Assistance

- [x] Yes (Please Specify the tool): **Claude Opus 5** (Claude Code)

## Discord username

Discord username: amogh_mavc

## Summary by CodeRabbit

- **New Features**
  - Pi telemetry extension setup is checked automatically after login.
  - The CLI can install, refresh, adopt, or migrate local Pi extensions, including backups for older installations.
  - Local installations record version information for status reporting.
  - Duplicate local extensions are removed when the npm-based extension is configured, preventing duplicate telemetry sessions.
  - Doctor commands report extension status and support safe cleanup and dry-run previews.

- **Documentation**
  - Added guidance for Pi extension installation modes, migration, backups, and unmanaged files.


Haz3-jolt (approved):
Thanks!

ShaanNarendran (changes requested):
Thanks for the thorough write-up and the honest mutation-testing notes — the test suite here is genuinely good, especially the negative cases around foreign files. Ran the five touched test files against the PR head (450 passed, ruff clean). A few things to sort before merge.

## Framing (please fix the PR title/description)

The description says the install path "shipped two defects that affect users upgrading". Nothing from #1602 has been merged — `observal_cli/pi_extension.py` doesn't exist on `main`, and the first commit here (`7d654ed9 feat(cli): auto-install Pi extension…`) *is* the feature. #1700 / #1720 were closed in favour of this one. So this is the #1602 implementation plus fixes to its own earlier revisions. The title should be `feat(cli): …` and the body should describe it as such; otherwise the changelog/reviewer context is wrong.

## Blocking

**1. Policy reversal not reflected in existing docs.**
Current `main` (`11b056d8`) treats the local bundled file as canonical and `npm:observal-pi` as *legacy* — `doctor patch` strips the npm registration. This PR inverts that (npm wins, local gets removed), which is what #1602 asks for, fine — but two docs still describe the old behaviour and weren't touched:
- `docs/cli/doctor.md:89` — "removes the legacy npm package registration"
- `docs/integrations/pi.md:104-108` — same claim

**2. Regression: content drift at the same version is never refreshed.**
On `main`, `_patch_pi` compares file content to the bundled source and rewrites on mismatch. Here, once the manifest says `managed: true` with `version == CLI version`, `check_status()` returns `CURRENT` without ever comparing content. Verified locally:

```
same-version content drift -> current None
```

Concretely: anyone on `uv tool install --editable .` editing `packages/pi-extension/extensions/observal.ts` gets "Already up to date" forever, since the version doesn't bump. Suggest: when the manifest version equals the current version, also check `installed == expected`; on mismatch return `action="refresh"` (with a `.bak`, same as migrate — the header check already implies overwriting edited copies is acceptable).

**3. `packages/pi-extension/README.md` — broken HTML comment.**
The new SPDX line is nested as `<!-- … -->` *inside* the existing `<!-- … -->` block. The inner `-->` closes the outer comment, so `SPDX-License-Identifier: Apache-2.0` and a stray `-->` will render as visible text on npm/GitHub. Drop the inner `<!--` / `-->`.

## Should-fix (non-blocking)

**4. `check_status()` runs up to three times per operation, and callers must pre-compute `backup_path()`.**
`_patch_pi` → `check_status()`, then `install_or_refresh()` → `check_status()` again, then a third call after dedupe. `_install_or_check_pi_extension` does the same double-call. CodeRabbit's race note is the symptom; this is the cause. Let `install_or_refresh()` accept an optional pre-computed `status` and return a small result (`action`, `backup: Path | None`) so the "read `backup_path()` before calling" footgun in the docstring goes away.

**5. `cleanup` doesn't keep a `.bak`, but the PR body says it does.**
`remove()` unlinks `MIGRATABLE` / `NPM_DUPLICATE` files with no backup. The body says "removed by `doctor patch` and `doctor cleanup`, keeping a `.bak`". Either add the backup to `remove()` for those two states, or correct the description and the README sentence ("doctor removes a local copy … keeping it as `observal.ts.bak`" reads as applying to both).

**6. Fifth copy of `_atomic_write`.**
`pi_extension._atomic_write` is byte-identical to `cmd_doctor._atomic_write`; `cmd_ops`, `cmd_pull`, `cmd_hook` have variants too. Since `cmd_doctor` now imports `pi_extension`, move the helper into `observal_cli/shared/utils.py` and have both import it.

## Nits

- `_is_observal_local_install` reads the manifest, then `check_status()` reads it again a few lines later in the local branch.
- FWIW I count 21 revisions of `observal.ts` on `main` carrying the signature, not 19 — all of them do, so the approach holds. The regression test guarding the header is a good call.
- Good to see the two vacuous tests from the body actually fixed in `36eea1b5`.

Requesting changes for #1–#3; the implementation itself is solid and I wouldn't hold it for #4–#6.

amogh-dongre:
@coderabbitai review the latest commit

ShaanNarendran (approved):

Haz3-jolt (approved):

ShaanNarendran (approved):

